PR #7, opened 2015-01-18
Added systemd service file.

PontusPersson:
Added a systemd service file to enable running systemd as a service on a system running systemd.


SimonPersson:
Thanks! Merged.

Issue #8, opened 2015-01-18
Spotifyd grabs control of audio when it's running


saevarb:
It seems that when spotifyd is running, it takes complete control of all sound. On my system, this means(for some reason) that youtube videos don't run at all(not even just on mute) while some videos(like those on facebook) play but without sound. Additionally, playing sound(from a video in chrome, for example) prevents spotifyd from starting properly until it seems to eventually crash with a floating point exception. This may be related to issue #2. 

I am running Arch Linux, and I'm going to take a wild guess that this has something to do with pulseaudio. 

I don't really know much about libalsa, so I'm unsure if this can be fixed without using another library.

SimonPersson:
I'll look into this on a machine running arch. Myself I run Crux and I can't reproduce this error, so pulseaudio seems like a reasonable guess.

Unfortunately I stole the libalsa code from the libspotify documentation, it seems like I will have to look into it after all.

PR #9, opened 2015-01-18
Fixes #2 & #8.

### LINKED ISSUE #2: Floating point exception (core dumped)
When I do this:

    $ sc QADDPL 5
    Added playlist "iTunes" to queue.

I get this:

    Floating point exception (core dumped)

saevarb:
The floating point exception people were running into is caused by the queue being empty and then trying next/prev which uses the modulus operator on `queue_len`, which is 0 if it is empty, resulting in division by zero.

The problem with spotifyd 'grabbing' audio turned out not to be a problem with spotifyd. See the commit message for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,15 @@ your own appkey from spotify if you don't trust me.
 Spotifyd requires Spotify Premium. This is a requirement on all applications using
 libspotify unfortunately.
 
+## Note for PulseAudio users
+spotifyd uses libalsa to play audio. If you are running a distribution that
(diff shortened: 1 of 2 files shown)


SimonPersson:
Thank you for investigating these errors. The fix looks good, but you used different indentation than I did. If you could make sure that the indentation is the same as the rest of the code, I'll merge it.

saevarb:
Ah yes, my bad. You're using tabs, I prefer spaces. Does that look good?

SimonPersson:
Yes, that looks good. Thank you very much for your kind words!

Merged.

PR #15, opened 2015-01-28
Updated spotifyd-nowplaying to notify if spotifyd isn't running.

saevarb:
EDIT: Yeah, I'm as dumb as a rock. I see you fixed the nowplaying script already. Anyhow, I added an extra line to deal with the case where spotifyd isn't running.

### DIFF
--- a/scripts/spotifyd-nowplaying
+++ b/scripts/spotifyd-nowplaying
@@ -32,6 +32,8 @@ output="$artist - $song"
 
 if [ "$scoutput" == "Not playing a track right now." ]; then
     echo $nothingplaying
+elif [[ $scoutput == "" ]]; then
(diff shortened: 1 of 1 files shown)


SimonPersson:
But, it already detects that it doesn't play, no? The check for not running is useful though.

Oh, now you updated the commit while I was writing. I'll merge that change! I try to test the scripts when there's updates to the output format so in hopefully you should just be able to take the latest script from the repo if things break :)

saevarb:
Yes, I'm sorry. I figured it wasn't your issue to fix the user-submitted scripts, so I didn't even check if it had been modified, and then worked from a local copy that didn't include the modifications from the repo.

Thanks!

PR #19, opened 2015-03-28
Add a command for setting the volume.

CMB:
vol takes a percentage from 0 to 100 and sets the volume to that
value.  The algorithm for adjusting the volume is probably too
simplistic, but it seems ok for me.


SimonPersson:
Could you also update the list of commands in the README?

I can't really say anything about the algorithm for lowering the volume. I'm not that concerned though, I imagine those that care about audio quality won't use this controller anyway.

Fix up the things I commented and I will merge it.

CMB:
Ok, updated the README and added braces.

SimonPersson:
Brilliant. Thank you for your contribution!

PR #26, opened 2015-04-15
Close alsa if no song is playing, fixes #16

### LINKED ISSUE #16: Close alsa device if no song is playing
Once you start playing some music using spotifyd and then stop the music by either pausing the playback or clearing the playlist it seems that the alsa device connection remains opened.

Because of that I have to for example restart spotifyd when I wake my device up from standby or
kill it for some applications that access /dev/snd directly(in my case forwarding alsa for docker containers).

I don't know if it's best practice or not but I think it would be appropriate for a daemon to close the
the alsa device when no song is currently playing.

As I am learning c at the moment I might be able to fix that issue myself very soon and open a pull request. What do you think about that issue?

Paspartout:
This small fix closes the alsa device if no audio data to play is available.
It automatically gets reopened when there is something to play.

I tried to make as less changes as possible and these changes seemed to be the
easiest way to achieve this. If you got any other idea please let me know!

### DIFF
--- a/src/alsa-audio.c
+++ b/src/alsa-audio.c
@@ -185,7 +185,7 @@ static void* alsa_audio_start(void *aux)
 	audio_fifo_data_t *afd;
 
 	for (;;) {
-		afd = audio_get(af);
+		afd = audio_get(af, &h);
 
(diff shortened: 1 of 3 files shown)


SimonPersson:
This seams like a nice solution. Thank you!

PR #36, opened 2017-01-12
Unbreak portaudio backend

dcoppa:
Fix Cargo.toml to unbreak librespot's portaudio backend


SimonPersson:
Nice catch! I made Travis build with portaudio and pulseaudio as well, so that it won't break in the future.

PR #58, opened 2017-06-02
Temp fix using updated librespot.

xmarinusx:
Had no experience whatsoever with Rust, so just hardcoded "Spotifyd" as device name to make compatible with new librespot version.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,11 +49,12 @@ struct MainLoopState {
     shutting_down: bool,
     cache: Option<Cache>,
     config: SessionConfig,
+    device_name: String,
     handle: Handle,
     discovery_stream: DiscoveryStream,
 }
 
-impl MainLoopState {
+impl MainLoopState {   
     fn new(connection: Box<Future<Item = Session, Error = io::Error>>,
            mixer: fn() -> Box<mixer::Mixer>,
            backend: fn(Option<String>) -> Box<Sink>,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -107,7 +107,7 @@ impl Future for MainLoopState {
                                          audio_filter,
                                          move || (backend)(audio_device));
 
-                let (spirc, spirc_task) = Spirc::new(session, player, (self.mixer)());
+                let (spirc, spirc_task) = Spirc::new("Spotifyd".to_string(), session, player, (self.mixer)());
                 self.spirc_task = Some(spirc_task);
```
SimonPersson: Perhaps we can add a `device_name: String` to `MainLoopState`, and then use it here instead of hard coding?

src/main.rs
```diff
@@ -211,7 +211,7 @@ fn main() {
     let backend = config.backend.clone();
     let audio_device = config.audio_device.clone();
     let device_id = session_config.device_id.clone();
-    let discovery_stream = discovery(&handle, session_config.name.clone(), device_id).unwrap();
+    let discovery_stream = discovery(&handle, "Spotifyd".to_string(), device_id).unwrap();
```
SimonPersson: Couldn't the hard coded string be replace by `config.device_name.clone()`?

xmarinusx:
This seems to do the trick.

SimonPersson:
Great! Thanks for your contribution 👍

PR #111, opened 2018-03-20
Edited README.md for correctness and clarity

vickychenglau:
I noticed some grammar and spelling errors in the README.md file and corrected them accordingly. I also made some changes for clarity.

Also, I would like to bring to attention the line:

> Values can be surrounded by double quotes (") which is useful if it contains the comment character (#).

was changed to

>Values can be surrounded by double quotes (") which is useful if **the value** contains the comment character (#).

for clarity. Please correct this is my assumption for what the "it" stood for was incorrect.


SimonPersson:
Thanks!

PR #122, opened 2018-04-07
Add description for cross-compilation

anduchs:
This adds some instructions on how to cross-compile an armhf version on ubuntu.
Tested this on xenial from a LinuxSubsystemForWindows.

I got those from my bash history. Before merge you might want to double-check.


SimonPersson:
Thanks for these notes on cross compilation. I think this might be better suited for the wiki. Do you mind adding it there instead? There are already some more detailed installation instructions there.

PR #128, opened 2018-04-23
Fix dbus property

smgt:
According to dbus specification it should be PlaybackStatus and not
PlayBackStatus.

https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html


SimonPersson:
Thanks!

Issue #133, opened 2018-05-17
i32 issue


wvangeit:
When compiling the latest master with 'cargo build --release', I get the error below.
I replaced i32 with u16 in both files, and at least the compilation finishes without error now.

   Compiling spotifyd v0.1.1 (file:///home/wvangeit/src/spotifyd)
error[E0308]: mismatched types
   --> src/main_loop.rs:147:33
    |
147 |                         volume: i32::from(mixer.volume()),
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^ expected u16, found i32

error[E0308]: mismatched types
  --> src/setup.rs:64:21
   |
64 |             volume: i32::from((mixer()).volume()),
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u16, found i32

error: aborting due to 2 previous errors

error: Could not compile `spotifyd`.

SimonPersson:
Is this after a `cargo update`? Updating the dependencies sometimes causes minor breakage. Do you mind making a pull request?

wvangeit:
Yes, that was after a cargo update.
I can make a PR, but I must say I'm basically working blind here because I don't know the code. Not sure if this code change can have undesired consequences.

praul:
I had this, also after cargo update. Changed the lines to u16, it compiles, and it works. Solves "called `Result::unwrap()` on an `Err` value: WireError(InvalidEnumValue(14))" for me

bphermansson:
Had this problem too and changing the files like @praul did worked for me.

SimonPersson:
This is fixed now, thanks @wvangeit

PR #135, opened 2018-05-17
Fix "expected u16, found i32" compilation error

wvangeit:
Fixes the compilation error as described in: https://github.com/Spotifyd/spotifyd/issues/133

As I mentioned in the ticket, I don't know the code so I'm not sure this code change doesn't have any undesired consequences.


SimonPersson:
You need to do a `cargo update` and add the `Cargo.lock` file, then it should work.

wvangeit:
Ok, thanks, that explains, Rust is new for me.

SimonPersson:
Thanks!

PR #141, opened 2018-05-25
Make the alsa backend optional and add support for the portaudio backend to enable OSX support

### LINKED ISSUE #129: Install on macOS
Is there any chance this library could be made to work on macOS?

Crazybus:
fixes #129 

I have no doubt there is a cleaner way to separate out all of the alsa logic without the need for conditional compilation. However I don't know any rust (yet) so this was just an attempt to prove that spotifyd will actually compile and run on OSX. If you have any tips on a cleaner way to implement this I would love some pointers in the right direction.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,6 +119,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bitflags"
+version = "0.3.3"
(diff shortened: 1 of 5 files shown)


SimonPersson:
Thanks for this!

I guess one possibility is to separate out the mixer logic into a separate crate in this repo, then we could contain all the conditional compilation logic there. I'll merge this until we have a cleaner solution.

Issue #165, opened 2018-09-03
Zombies attack after events


ah-:
First off, thanks for building spotifyd, it's amazing!

I'm using it on a small arm sbc and have it turn my stereo on/off automatically via a wifi power socket, which is triggered via spotifyd events:

```
#!/bin/bash

case "$PLAYER_EVENT" in
"start")
    curl "http://sonoff/cm?cmnd=Power%20On"
    ;;
"stop")
    curl "http://sonoff/cm?cmnd=Power%20Off"
    ;;
esac
```

However, after a while this seems to leave behind quite a few defunct processes:
```
root     14596  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14599  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14600  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14601  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14602  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14604  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14607  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14608  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14747  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
root     14776  0.0  0.0      0     0 ?        Z    Sep01   0:00 [spotify-event.s] <defunct>
```

(Yes, I know I shouldn't run it as root, i'll get around to that).

Any idea what's going on? In the logs it all looks fine:
```
Sep 02 23:52:32 alarm spotifyd[12372]: {"POWER":"OFF"}23:52:32 [INFO] Running ["/home/andreas/spotify-event.sh"] with environment variables {"TRACK_ID": "212guZJYmzASElwy8HmVKx", "PLAYER_EVENT": "start"}
Sep 02 23:52:32 alarm spotifyd[12372]:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[...]

PR #166, opened 2018-09-04
Wait for event programs to exit

ah-:
Fixes https://github.com/Spotifyd/spotifyd/issues/165

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1,6 +1,7 @@
 use futures::{Async, Future, Poll, Stream};
 use futures;
 use std::io;
+use std::process::Child;
 use std::rc::Rc;
 
 use librespot::connect::spirc::{Spirc, SpircTask};
@@ -92,6 +93,7 @@ pub struct MainLoopState {
     pub session_config: SessionConfig,
     pub handle: Handle,
     pub linear_volume: bool,
+    pub running_event_program: Option<Child>,
 }
 
 impl Future for MainLoopState {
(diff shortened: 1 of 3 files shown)


SimonPersson:
Nice catch!

So the reason why I didn't add a wait statement was because I figured that the user might run a process that takes some time to finish, which would introduce little lag on each event.

Maybe we can work around that by running the processes from a separate thread? Something like the global threadpool in rayon [1] should work.

[1] https://docs.rs/rayon/*/rayon/struct.ThreadPool.html#method.spawn

ah-:
Yep, I see that. Actually happened to me as well when the HTTP requests took a while to go through successfully, then it just hangs the main loop which isn't great.

Looking a this again, we have two other options as well: We could set `signal(SIGCHLD, SIG_IGN);` and just ignore all children (but I'm not sure if that would affect anything else), or we could keep the child processes in a list and use https://doc.rust-lang.org/stable/std/process/struct.Child.html#method.try_wait to reap them.

Both would need some testing as I'm not yet quite sure how they integrate with `Command`, and I'm away for a few days now but will have another look when I'm back.

hajdbo:
this PR should be merged.

SimonPersson:
@ah- Yes, `try_wait` sounds right. Do you want to try to implement it? Storing the pending processes in `MainLoopState` is fine by me.

ah-:
Yes, updated the PR with an initial implementation of `try_wait`.

I've gone with only running one script at a time, which is useful anyways to ensure there are no race conditions.

The only thing I'm not yet quite sure about is if the integration with the main loop future is perfectly correct. There's currently nothing to wake the future when the process exits, it relies on other wakeups to advance.

src/main_loop.rs
```diff
@@ -113,10 +115,21 @@ impl Future for MainLoopState {
                     Session::connect(session_config, creds, cache, handle);
             }
 
-            if let Some(ref mut player_event_channel) = self.spotifyd_state.player_event_channel {
-                if let Async::Ready(Some(event)) = player_event_channel.poll().unwrap() {
-                    if let Some(ref program) = self.spotifyd_state.player_event_program {
-                        run_program_on_events(event, program);
```
ah-: Maybe there's a nice way of expressing the next block, all I want is to set the Option to None if try_wait is successful. With NLL I could rewrite this to be a bit more straight forward, but on stable I haven't seen anything better than taking the value out of the Option and then putting it back.

PR #174, opened 2018-10-05
Updated Cargo.lock version and added ap_port field required by librespot.core.config.SessionConfig

codyshepherd:
Running `cargo update` seems to have included `librespot`, which has added another key value pair to core.config.SessionConfig.

I've built and run this successfully on Ubuntu Cosmic (18.10).


SimonPersson:
Thanks!

PR #188, opened 2018-12-12
Add dependencies to the spotifyd systemd service

bertptrs:
Spotifyd crashes if it starts before either the sound card is configured or the network is available. Whilst this is not too bad, it causes a restart of the service when started at boot. It is nicer to just depend on the targets that systemd provides to signal availability of both.

### DIFF
--- a/contrib/spotifyd.service
+++ b/contrib/spotifyd.service
@@ -1,6 +1,10 @@
 [Unit]
 Description=A spotify playing daemon
 Documentation=https://github.com/Spotifyd/spotifyd
+Wants=sound.target
+After=sound.target
+Wants=network-online.target
+After=network-online.target
 
 [Service]
 ExecStart=/usr/bin/spotifyd --no-daemon


NicoHood:
I am not sure, but wants and after at the same time looks wrong to me.

bertptrs:
`Wants` and `After` together is not weird, but actually very common since they are complementary. You can grep your unit files for this.

`Wants` is a soft-dependency. It pulls the unit in if it wasn't going to be started yet, but contrary to a `Requires`, it doesn't actually prevent the unit from starting if its dependency has failed. This can be useful since some targets tend to partially fail, but still work. `network.target` on Ubuntu is notorious for this in my experience.

However, `Wants` on its own does not establish a critical path; it simply adds the service to the list of services to be started, so the dependency can be started at the same time as the service itself. This is what `After` is for; it says that a particular service needs to be done starting before this one can begin. `After` does not pull in the dependency if it wasn't already going to be started, it just make sure of the ordering.

Both have their uses, and both _can_ be used separately, but in this case you really want both. You can read more about this [in the man page](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Requires=).

A case can be made for the `Wants=` to be turned into `Requires=`, but I think the potential downsides outweigh the upsides.

SimonPersson:
Thanks!

PR #189, opened 2018-12-13
Use port 443 instead of 80

Jalle19:
Some networks block Spotify Connect, probably using some simple packet inspection. Switching to port 443 fixes this, at least for me.


SimonPersson:
Thanks!

PR #194, opened 2019-01-03
Use config in /etc for systemd service

NicoHood:
Even though it is a user service, it is still recommended to load the configuration from a central configuration place. A configuration file is a must have for a service, you should not edit the service file itself.

I am also working on a system wide service that works on embedded systems with no user logged in. I am trying to get this working properly first, so it feels better integrated and not so 'hacky'. The user service is the correct option for now, especially when running a normal desktop environment at the same time.

### DIFF
--- a/contrib/spotifyd.service
+++ b/contrib/spotifyd.service
@@ -7,7 +7,7 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-ExecStart=/usr/bin/spotifyd --no-daemon
+ExecStart=/usr/bin/spotifyd --no-daemon --config /etc/spotifyd.conf
 Restart=always
 RestartSec=12
 


SimonPersson:
I'm not sure I understand your justification for this change. Spotifyd will already look up the config file in `/etc/spotifyd.conf`, although you can override it by creating a file in the home directory. Is that not the expected behaviour?

NicoHood:
Oh, I missed that. In which order are configs loaded from where?

SimonPersson:
Local configuration gets read first, where exactly might depend on your XDG configuration, but for me it's at `~/.config/spotifyd/spotifyd.conf`. If that file is not found, `/etc/spotifyd.conf` is read instead. This means that you can have a global default configuration, and that users can override it locally.

NicoHood:
Alright, then my PR is not required, thanks.

It would be nice if you can provide a sample config as file in the repository (not only the readme), along with some more detailed explanation and comments. This could then be packaged better.

PR #195, opened 2019-01-06
Fix --device_name argument parsing

ismacasti:
The device_name command parameter was not being passed
to the main loop.


SimonPersson:
Thanks!

PR #199, opened 2019-01-19
Replace rust-crypto with sha-1 crate

newpavlov:
You don't need the whole rust-crypto for just SHA-1. Also see librespot-org/librespot#239.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1929,7 +1929,7 @@ dependencies = [
 [[package]]
 name = "rust-crypto"
 version = "0.2.36"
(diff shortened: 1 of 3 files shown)


bertptrs:
Any comments on this PR? The `rust-crypto` crate has been all but abandoned, while the [RustCrypto](https://github.com/RustCrypto) group (that maintains this and quite a few other hashes) is very active. Plus, now that librespot also includes the `sha-1` crate over the `rust-crypto` crate, there is some overhead in including both libraries.

SimonPersson:
Sorry for the delay. I was actually waiting for librespot to merge this first, to avoid both crates. Thanks for bringing it to my attention again!

PR #206, opened 2019-02-22
Bump librespot version.

### LINKED ISSUE #205: Can't change playlist
Today I discovered an issue using spotifyd daemon: I can't play music that is not in the current playlist. So it's not possible to switch to another playlist or to play music that was searched.  This happens when I control the daemon both with an android or windows version of Spotify app. This behaviour only occurs with the spotifyd daemon not when I use the remote functionality to control 'no-linux' devices. I installed the latest version of spotifyd daemon with the same result. Has Spotify changed something that affects spotifyd?

There is a workaround for this by going back to the device, play music there, change the playlist and then go back to spotifyd again, but this is very annoying.

aksel:
Resolves #205 

I am unsure if I have gone about this the wrong way; haven't worked with Rust before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -793,7 +793,7 @@ dependencies = [
 [[package]]
 name = "librespot"
 version = "0.1.0"
-source = "git+https://github.com/librespot-org/librespot.git#d2cadec4190a44572b1d5d5daed4f0eda1c2b921"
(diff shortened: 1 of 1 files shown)


woftor:
Tested: works.

zetorian:
Works for me as well, no more issues after suspend / hibernate

SimonPersson:
Thanks for this!

PR #210, opened 2019-03-02
Add single artist to mpris metadata

christopher-dG:
This adds the `xesam:artist` metadata field from the first artist, which is usually the only one.


SimonPersson:
Thanks!

PR #217, opened 2019-03-16
Add version flag

### LINKED ISSUE #208: '--version'?
Is there a way to ascertain the version of a `spotifyd` binary (besides consulting sizes or checksums from the releases page?) If not, a `--version` flag seems like it would be handy for this.

(Also, it looks like the configured Cargo package version isn't up to date w/ the version number of the release? E.g. release 0.2.4: https://github.com/Spotifyd/spotifyd/blob/2abc39285d95481d5ae359a5d641c08663cd4f02/Cargo.toml#L3 )

Thanks for developing, sharing, maintaining and supporting `spotifyd`. I use `spotifyd` every day and am exceptionally happy w/ its utility and performance.

SirWindfield:
This PR adds a `--version, -V` flag to the cli. It returns the version specified in the `Cargo.toml` file. The value is read at compile time and embedded into the executable. This is done by using a new macro called `crate_version!` (shamefully taken from [clap](https://github.com/clap-rs/clap), a great command line parser).

I also updated the version number to `0.2.6` as the latest release was `0.2.5`. The `Cargo.toml` file was still specifying version `0.2.1`.

Closes #208.


SimonPersson:
Thanks!

PR #221, opened 2019-03-22
Move to 2018 edition

SirWindfield:
This PR moves partial source code to the 2018 edition of Rust.

API wise, this won't change much. The 2018 edition improved the way how Rust code gets imported. Instead of having to write a huge chunk of `extern crate ...` statements, the code gets now pulled in as soon as someone declares its usage via `use ...`. The change introduced a new keyword too, `crate`, that is used to refer to the crate's own modules.

To make use of the 2018edition I had to upgrade the `syslog` and `log` crates. Those changed their API from `LogLevelFilter` to `LevelFilter` and I changed those two lines accordingly. 

Besides that, the 2018 edition introduced a new operator, the `?`. This one basically is the old `try!` macro you used in some part of the code. The team behind Rust decided to give it its own operator as it has been used extensively by the community and them.

I've build every feature available locally to make sure that everything still compiles. I didn't change any logic whatsoever in this PR so the daemon will work as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,9 @@ dependencies = [
 name = "error-chain"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace 0.3.13 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


SimonPersson:
Thanks!

Issue #163, opened 2018-08-20
Spotify access token instead of plain text credentials


Thierry-Gls:
Hi,

I was wondering if there is a possibility to implement use of access token to connect a spotify account to spotifyd, because storing plain text credentials datas can be a really huge security gap.

I've already seen in the issue #148 that it might be difficult to change the way it works, but instead of using a hashed password, is this not possible to use the access token from spotify web api ?

SimonPersson:
The spotify web api does not allow for streaming/playback, so I don't think access tokens from there are useful for our purposes. I think the keychain proposal in #148 is the way forward.

HowToMeetLadies:
No need for personal credentials, tokens or whatever:
https://www.spotify.com/account/set-device-password/

NicoHood:
This link seems to not work for me. What is displayed there?

tjorim:
That generates a device password. However, this is only for Spotify accounts opened with a Facebook account. Still, that are still plain text credentials, to be used when you delete or lose access to your Facebook account.

SirWindfield:
Ok, looks like you have to declare each close separately in a PR. This can be closed now, as keyring support has been added with the latest commit.

PR #223, opened 2019-03-24
Add keyring password support

### LINKED ISSUE #148: Plain text password
The password is stored in plain text in the config file. Is there any chance to store a hashed version of the password?

I also tried to run without a user/password and it works perfect. This might be only working in the local network. No matter why, you should add this to the readme, which is an extremely nice feature. You write "Every field is optional; Spotifyd can even run without a configuration file.", but I was not aware that this means no authentification is required at all. Even more: two different accounts can stream to the same spotifyd, which is extremely great for us, as everyone can use the kitchen speakers now, no need to login with a single account.

Great software, thanks for sharing!

SirWindfield:
This PR adds support for the Linux Secret Service API that can be used to retrieve passwords from various keychains.

It adds an additional CLI flag and config entry called `use-keyring` that, if set to `true`, tries to retrieve the password from the keyring if none has been specified in the config file. 

The keyring entry has to have the following attributes set:
```
application: rust-keyring
service: spotifyd
username: <spotify-username>
```
Here is an example on how to store the password using `secret-tool`, a CLI that can communicate with almost every keyring over dbus:
```
$ secret-tool --label='entry name that you can choose' application rust-keyring service spotifyd username <your-username>
```

`spotifyd` will only check the keyring if the boolean flag has been set to `true` and a Spotify user name has been configured.

The underlying library used is `keyring-rs`, the one mentioned in #148.

This PR has been locked behind a feature flag as I am not entirely sure if every distribution has access to the DBUS and therefore to the underlying API.

Closes #148, #163.


SimonPersson:
Cool! Could you add some documentation to the README for this feature?

SirWindfield:
Haha what a timing! I just remembered that I forgot to add some description to the README file.

SimonPersson:
Great, thanks!

PR #225, opened 2019-03-25
Refactor code

SirWindfield:
This PR fixed a lot of linting warnings that were outputted by [clippy](https://github.com/rust-lang/rust-clippy).

Some of them were possible bugs (like the number conversions used for the volume of the mixer), some were just shorter code.
I had to add an exception for clippy to not report a line 79 in `setup.rs` as the value gets moved and it would be impossible (with the current code) to access the `config` value within a closure.

Most of the warnings where related to the structure shorthand notation. If the variable name and the struct field name are the same, the value can be passed by just writing `field_name` instead of `field_name: field_name`. 
Another common one was the usage or `*_or` instead of `*_or_else`. The latter is preferred as the closure that gets passed to it will only get executed if it actually reaches that part of the chain.

Additionally I've added a CI `before_script` entry to install `clippy` and an entry to `script` to check for any warnings. The CI will fail if the code does contain one.
I also pinned down the rust-toolchain version down to the one from yesterday, as it is the newest nightly version that compiles with `clippy` and `rustfmt`, a tool used to format source code.
I was looking into adding support for `rustfmt` too.

Let me know if you would be OK to introduce the linting tool. It prevents common mistakes and improves the code quality. Additionally I would love to add [rustfmt](https://github.com/rust-lang/rustfmt), the official formatting tool. Both of these would make sure that the code pushed to Github consists of a standardized styling and has a set baseline when it comes to code quality. I would configure it to mirror your coding style used throughout the current code base.


SimonPersson:
Amazing! Feel free to introduce rustfmt. We can go with the default options, as I don't really mind and I think it helps for collaboration.

PR #227, opened 2019-03-28
Add missing `store` keyword in secret-tool command example

joelpet:
Previously, the command would fail with error code 2 and display the following
usage description:

```
usage: secret-tool store --label='label' attribute value ...
       secret-tool lookup attribute value ...
       secret-tool clear attribute value ...
       secret-tool search [--all] [--unlock] attribute value ...
```


SimonPersson:
Thanks!

PR #228, opened 2019-03-28
Clarify usage of the use-keyring config option

joelpet:
Previously, there was no clear indication that a username must be explicitly
provided to Spotifyd rather than looked up in the keyring.


SimonPersson:
Great, thanks!

PR #245, opened 2019-05-03
Update mpris metadata

pstruschka:
Align Spotifyd's metadata fields with specification guidelines and
Spotify's metadata fields, for compatibility with existing scripts that
make use of mpris metadata.

Unified formatting for setting the metadata fields.


SimonPersson:
Great, thanks!

Issue #87, opened 2017-11-12
Alsa errors but sound works


sebastianst:
Even though sound works, the logs show the following errors:
```
19:09:37 [INFO] Using alsa sink
ALSA lib conf.c:4884:(parse_args) Unknown parameter 1
ALSA lib conf.c:5017:(snd_config_expand) Parse arguments error: No such file or directory
ALSA lib control.c:1373:(snd_ctl_open_noupdate) Invalid CTL hw:HDMI,3
19:09:37 [ERROR] Couldn't set volume of alsa device.
```
Relevant parts from config:
```
backend = alsa
device = hw:HDMI,3 # Or change according to `aplay -L` or `aplay -l`
volume-control = alsa # or softvol
```
Alsa devices:
```
# aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: HDMI [HDA ATI HDMI], device 3: HDMI 0 [HDMI 0]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: HDMI [HDA ATI HDMI], device 7: HDMI 1 [HDMI 1]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: Generic [HD-Audio Generic], device 0: ALC671 Analog [ALC671 Analog]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: Generic [HD-Audio Generic], device 1: ALC671 Digital [ALC671 Digital]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```
Alsa PCMs:
```
# aplay -L
null
    Discard all samples (playback) or generate zero samples (capture)
hdmi:CARD=HDMI,DEV=0
    HDA ATI HDMI, HDMI 0
    HDMI Audio Output
hdmi:CARD=HDMI,DEV=1
    HDA ATI HDMI, HDMI 1
    HDMI Audio Output
default:CARD=Generic
    HD-Audio Generic, ALC671 Analog
    Default Audio Device
sysdefault:CARD=Generic
    HD-Audio Generic, ALC671 Analog
    Default Audio Device
front:CARD=Generic,DEV=0
    HD-Audio Generic, ALC671 Analog
    Front speakers
surround21:CARD=Generic,DEV=0
    HD-Audio Generic, ALC671 Analog
    2.1 Surround output to Front and Subwoofer speakers
surround40:CARD=Generic,DEV=0
    HD-Audio Generic, ALC671 Analog
[...]

PR #254, opened 2019-06-15
Separate config option for control (i.e. mixer) device

eisnerd:
See #87.

I'm very happy to change option name, description, etc. and be corrected on my rusty rust. I just found this useful and wanted to share.


SimonPersson:
Thanks!

PR #257, opened 2019-06-22
Enable clippy, rustfmt and pulseaudio for x86

Mitschmaster:
I reenabled clippy and rustfmt. i had no success in enabling pulseaudio for both architectures. so i left it only for x86.


SimonPersson:
This is great, thank you!

PR #267, opened 2019-07-03
Improve english on cache_path explanation

RlndVt:
Define `the cache path' and remove ambiguity, 

Remove the repetitive 'is set to A so A is used'.

`have to exist' is not english. The bracketed text can also be removed because an explanation that the path must exist follows soon after.

First explain missing path and then shell expansion.

Remove redundant 'this means that ...'.

As a followup, what would happen if a relative path is provided? Is it allowed or must the path be absolute?

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,12 @@ the `[global]` section. This is useful when you run applications related to
 Values can be surrounded by double quotes (") which is useful if the value contains
 the comment character (#).
 
-The line `cache_path = /cache_directory` defines where Spotify's cache-files are stored.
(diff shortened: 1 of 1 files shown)


SimonPersson:
Thanks! Relative paths are allowed. Do you think that needs further specification?

RlndVt:
Yes absolutely.

What is it relative to?

SimonPersson:
To the working directory at the time you executed spotifyd.

RlndVt:
... that makes sense :)

I'll see if I can add that to the readme.

Might not need to in the end.

PR #283, opened 2019-09-08
merge cli and config flags

### LINKED ISSUE #143: Enable http proxy configuration
Currently, spotifyd doesn't work with `https_proxy` setting.

https://github.com/Spotifyd/spotifyd/issues/129#issuecomment-392869695 by @tnguyen14
>```[WARN] Failed to register IPv6 receiver: Os { code: 49, kind: AddrNotAvailable, message: "Can\'t assign requested address" }```
> It is likely due to the fact that I have a work VPN on. I do have https_proxy set to go to the Internet. Do you know if Spotifyd would work with proxy?

and https://github.com/Spotifyd/spotifyd/issues/129#issuecomment-392871871 by @SimonPersson 

>@tnguyen14 - We don't have proxy support, but it would be simple to add. See https://github.com/Spotifyd/spotifyd/blob/master/src/config.rs if you want to hack on it. Essentially we just need to add it as a config option and pass it along to librespot.

### LINKED ISSUE #233: Unify config file key names
As of now, some config file key names have different styles, e.g. underscore vs. hyphen.

### LINKED ISSUE #265: Support for zeroconf flag in librespot
When using a firewall, the spotifyd client cannot be found by other devices. 
librespot allows using a [flag that sets a port value](https://github.com/librespot-org/librespot/issues/91).

It would be nice to have this flag in spotifyd too.

SirWindfield:
This PR merges the CLI and config options into one common struct to prevent different naming schemes. As far as I can tell, the change is fully backwards compatible thanks to clap's and serde's  `alias` keywords.

I basically only changed the parsing logic, the outer interface is still the same (`SpotifydConfig`).

CLI arguments take priority over config arguments. They are merged together during load. In theory, different file formats other than INI could be supported but I opted out of this one for now. I had to re-declare `librespot` enums like `Bitrate` to add help text to them.

The following keys have changed (the old keys could be removed in a breaking change later on):

| Old value             | New Value             |
|-----------------------|-----------------------|
| volume-control        | volume_controller     |
| volume-normalisation  | volume_normalisation  |
| use-keyring           | use_keyring           |
| normalisation-pregain | normalisation_pregain |
| onevent               | on-song-change-hook   |

The idea behind the last one was to add more unified hooks later on (like `on-login-hook` or `on-device-connected-hook`, meaning that people can create more custom scripts to show notifications or do other stuff).

I still need to clean up the code as I can re-arrange a lot of stuff to make it more readable.
Creating the PR so you guys know that someone is working on it :)

The following things still need to be done:

- [x] cleanup dependencies
- [x] re-format code
- [x] re-structure cli and config structs (embedding them the other way around makes more sense)
- [x] implement zeroconf and http 

The PR will, after completion, affect the following issues:
Closes #233.
Closes #265.


SirWindfield:
I think I got everything done. Would be nice if you could take a look and give an OK :)
The flags and the config entries should be completely backwards compatible.

SimonPersson:
Thank you for this! I have been away from my computer for some time, so I wasn't able to review it for you. Looks like a great improvement :)

PR #317, opened 2019-09-22
add CI and CD workflows

### LINKED ISSUE #281: Provide macOS binaries
This can be actually done using Travis CI.

SirWindfield:
This commit adds a CI workflow to the project. It checks for the correct code style (using cargo fmt) and lints the whole project as well (using cargo clippy). The last step checks if the binary can be compiled on our target OS (using cargo check).

The commit also adds a workflow for CD. After each release, the workflow will build binaries for both Linux and macOS in two flavours: slim and full. Slim does contain nothing but the proper audio backend. Full contains all features that work correctly on the specific platform. For Linux that is dbus_keyring and dbus_mpris. For macOS it is only dbus_keyring. The resulting binaries are compressed into a .tar.gz archive and uploaded as assets to the appropiate release. The script does also create sha512 checksums and uploads them as well.

Closes #281.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,68 @@
+name: Continuous Deployment
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  publish:
+    name: Publishing for ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+        rust: [stable]
+        feature: [
+          '',                                             # Slim version has no features enabled by default.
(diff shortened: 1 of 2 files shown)


SimonPersson:
This looks really nice. It looks like the CD pipeline only makes builds for AMD64 architectures, right? Is it possible to support ARM as well? Or is the plan to keep using Travis CI, in parallel with Github Actions?

SirWindfield:
I still need to figure out a decent way to distinguish between the runtime that compiles for amd64 and the one that compiles to arm. Maybe a simple matrix entry like "arch" would do the trick so I can check during the dependency installation step which arch the target should be compiled for and install the appropiate armhf dependencies.

In the long run yes, it would be great to have it all in Github Actions :) But I only got the two (kind of) working.

Issue #297, opened 2019-09-18
Make CLI flags feature aware


SirWindfield:
As of now, the CLI's help page always displays all flags, regardless of whether they are applicable or not. For example, if the `dbus_keyring` feature has not been specified, the CLI still shows the flag. This happens with the possible values for `backend` as well.

jackloughran:
@SirWindfield I didn't use the correct closing issue keyword in my pr, so this didn't close.

SirWindfield:
Oh yeah. Thanks again!

PR #338, opened 2019-09-27
feature aware cli flags

jackloughran:
An attempt at resolving #297 

This first commit seems to filter the backends okay, but I'm having a bit of trouble figuring out the best way to conditionally show the `use-keyring`. I can't think of a way to conditionally toggle a field since it's specified with the structopt attribute. I'm still relatively new to rust, so if anyone could point me in the right direction, that'd be great!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -197,7 +197,7 @@ pub struct SharedConfigValues {
     password: Option<String>,
 
     /// Enables keyring password access
-    #[structopt(long)]
+    #[cfg_attr(feature = "dbus_keyring", structopt(long))]
     #[serde(alias = "use-keyring", default, deserialize_with = "de_from_str")]
     use_keyring: bool,
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -207,8 +207,11 @@ pub struct SharedConfigValues {
     password: Option<String>,
 
     /// Enables keyring password access
-    #[cfg_attr(feature = "dbus_keyring", structopt(long))]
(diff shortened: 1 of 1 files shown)


jackloughran:
@SirWindfield Thanks for the help! While looking into your suggestion, I found `cfg_attr`, which seems to do the job pretty well.

As for the volume controller, it looks like you're right and the alsa volume controllers are only available with `alsa_backend` based on https://github.com/Spotifyd/spotifyd/blob/fc6ed885eac6b512b6e036fdb8dbfa327f9d283a/src/setup.rs#L63-L68

SirWindfield:
Make sure to run `cargo fmt`. The code style check failed. :)

src/config.rs
```diff
@@ -185,7 +207,7 @@ pub struct SharedConfigValues {
     password: Option<String>,
 
     /// Enables keyring password access
-    #[structopt(long)]
+    #[cfg_attr(feature = "dbus_keyring", structopt(long))]
     #[serde(alias = "use-keyring", default, deserialize_with = "de_from_str")]
```
SirWindfield: Does it make sense to add the cfg_attr to the serde field too? In theory it wouldn’t change anything as the check only happens if the feature I set anyway. But it would be more consistent behavior as the field can’t be modified in any way if the feature is not available

jackloughran: Yeah, that makes sense.

SirWindfield:
Thanks for your contribution!

Issue #353, opened 2019-10-03
MPRIS PlayerStatus always returning "Stopped"


SirWindfield:
It looks like the Connect API returns the SHA-1 encoded device_name instead of the raw one. We compare it to the unhashed version. And since that will always fail, the PlayerStatus will always be Stopped.

I am, however, sure that it worked some releases ago 🤔 

Relevant lines:
https://github.com/Spotifyd/spotifyd/blob/6c2fec5fef28c3565a467769d17f7ceec43e91bb/src/dbus_mpris.rs#L344
`player.device.id` is the hash, `_device.unwrap()` is the URL-encoded device name.

https://github.com/Spotifyd/spotifyd/blob/6c2fec5fef28c3565a467769d17f7ceec43e91bb/src/config.rs#L451

To implement this, the above method should probably be moved to `utils.rs`

stuarth:
I'll take this one!

hifiberry:
Just had a look and it seems it also doesn't work on 0.2.10 anymore as it was some days ago. Might  be a change in the API.

SirWindfield:
That was my guess too. The online API does mention the hash though but I didn't check Google cache or the web archive for an older version.

SirWindfield:
Just checked archive.org's fetches of the API page but even in 2017 the device_id is the hash of the device's name. Looks like my brain is tricking me...

SirWindfield:
@stuarth FYI, https://github.com/ramsayleung/rspotify/blob/fa0408d3278aa7f87fe6fe8e30f88b785cdea20c/src/spotify/model/device.rs#L5 is the struct returned by `player.device`. It has a name field. So it might be enough to compare to `player.device.name` instead of `player.device.id`.

stuarth:
@SirWindfield Thanks! Looking into it now

PR #356, opened 2019-10-06
fix dbus-mpris device name comparison

stuarth:
The bug is exactly as mentioned by @SirWindfield [on the issue](https://github.com/Spotifyd/spotifyd/issues/353#issuecomment-538642273). Fix is to compare against the URL-encoded device name.

Fixes https://github.com/Spotifyd/spotifyd/issues/353.


SirWindfield:
Thank you very much!

PR #357, opened 2019-10-06
Added option to use credential-only caching as described in #351

### LINKED ISSUE #351: Allow credential-only caching
`librespot` allows to differentiate between audio and credential caching or credential-only caching. We could introduce a new config entry and pass it down to `librespot`.

https://github.com/Spotifyd/spotifyd/blob/5d6771c60d9cfecc1030c6514ee69037371c4227/src/config.rs#L483

`true` indicates the Cache to also cache audio files. Implementing this feature would mean passing down a bool and adding a new field to the configuration struct instead of hard-coding it to `true`.

TobiMichael96:
Added option to use credential-only caching as described in issue #351 

By default this uses credential-only caching which might be not the best approach, as the default value was hardcoded with "true".

Let me know if you want to change this.

Disclaimer: I'm pretty new to all of this, so just tell me if I did something stupid. 

Edit: Fix #351

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -235,6 +235,11 @@ pub struct SharedConfigValues {
     #[structopt(long, parse(from_os_str), short, value_name = "string")]
     cache_path: Option<PathBuf>,
 
+    /// Enable to use audio cache
+    #[structopt(long)]
+    #[serde(default, deserialize_with = "de_from_str")]
+    audio_cache: bool,
+
     /// The audio backend to use
     #[structopt(long, short, possible_values = &BACKEND_VALUES, value_name = "string")]
     backend: Option<Backend>,
@@ -341,6 +346,7 @@ impl fmt::Debug for SharedConfigValues {
             .field("use_keyring", &self.use_keyring)
             .field("on_song_change_hook", &self.on_song_change_hook)
             .field("cache_path", &self.cache_path)
+            .field("audio_cache", &self.audio_cache)
             .field("backend", &self.backend)
             .field("volume_controller", &self.volume_controller)
             .field("device", &self.device)
@@ -431,6 +437,7 @@ impl SharedConfigValues {
         // Handles boolean merging.
         self.use_keyring |= other.use_keyring;
         self.volume_normalisation |= other.volume_normalisation;
+        self.audio_cache |= other.audio_cache;
     }
 }
 
@@ -458,6 +465,8 @@ pub(crate) struct SpotifydConfig {
     #[allow(unused)]
     pub(crate) use_keyring: bool,
     pub(crate) cache: Option<Cache>,
+    #[allow(unused)]
+    pub(crate) audio_cache: bool,
     pub(crate) backend: Option<String>,
     pub(crate) audio_device: Option<String>,
     #[allow(unused)]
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -542,6 +553,7 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
         password,
         use_keyring: config.shared_config.use_keyring,
         cache,
+        audio_cache,
```
SirWindfield: The SpotifydStruct is only used internally. And since no other methods rely on the audio_cache flag, this can be removed.

src/config.rs
```diff
@@ -235,6 +235,11 @@ pub struct SharedConfigValues {
     #[structopt(long, parse(from_os_str), short, value_name = "string")]
     cache_path: Option<PathBuf>,
 
+    /// Enable to use audio cache
+    #[structopt(long)]
+    #[serde(default, deserialize_with = "de_from_str")]
+    audio_cache: bool,
```
SirWindfield: I think I would rename it to cache_audio. Makes the command line more “readable”:
`spotifyd —no-daemon —cache-audio —use-keyring`

It also matches more the naming scheme and makes reading the line more natural.

SirWindfield: Oh wait nevermind. If we use true as the default the flag should probably be no-audio-cache instead. My bad!

SirWindfield:
For future reference, take a look at [this](https://help.github.com/en/articles/closing-issues-using-keywords). You can edit your first comment and github will still recognize it :)

TobiMichael96:
> For future reference, take a look at [this](https://help.github.com/en/articles/closing-issues-using-keywords). You can edit your first comment and github will still recognize it :)
> 
> I really need to get the issue and PR templates going. Would make it way easier for others to contribute :)

Thank you, I'll use this in the future!

TobiMichael96:
> To make things easier and not bump for a minor version change I would rather make the default value true.
> Oh, and thanks for spending some time on Spotifyd :)

Alright, I change the things as you said, I hope it's ok now.

No need to thank me, I'm using it on a daily basis, so I wanted to give something back (even if it's pretty small compared to what you do)!

src/config.rs
```diff
@@ -476,11 +483,13 @@ pub(crate) struct SpotifydConfig {
 }
 
 pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
+    let no_audio_cache = !config.shared_config.no_audio_cache;
```
SirWindfield: Last request :) Would you mind renaming `let no_audio_cache` to `let audio_cache`? Makes more sense since the `no_audio_cache` negated determines if the `audio_cache` is used.

SirWindfield: If CI passes afterwards I'll merge it :smile:

TobiMichael96: Sure thing! :)

TobiMichael96:
Maybe we should add the option in the README like this:

```ini
# ~/.config/spotifyd/spotifyd.conf
[global]
volume_controller = alsa                                    # or alsa_linear, or softvol
# on_song_change_hook = command_to_run_on_playback_events
device_name = device_name_in_spotify_connect             # must not contain spaces
bitrate = 160                                            # or 96, or 320
cache_path = cache_directory
**no_audio_cache = true                           # to use credential-only caching**
volume_normalisation = true
normalisation_pregain = -10
```

SirWindfield:
Probably ye :)

Issue #355, opened 2019-10-05
Add `debug` CLI config that prevents logging from taking out values for privacy


SirWindfield:
This would be a nice to have feature to make debugging `BadCredentials` during the login step of the daemon easier to debug. It would allow people to see their password and username used within the configuration.

How to implement:
- Add a boolean flag (name is up for discussion) to the `SharedConfigValues` struct.
- Check for the flag [in this method](https://github.com/Spotifyd/spotifyd/blob/ea66c13a0288c0b8ec39c857fcf5b354ceb070b0/src/config.rs#L315) and change the values accordingly.

olavbm:
I'd like to contribute here.

SirWindfield:
Someone created a PR linked above (#360). We came to the conclusion that it would be better to move our Debug implementation one scope up in the chain. See https://github.com/Spotifyd/spotifyd/issues/360#issuecomment-539150809 for more information. 

I am not sure how much experience you have with Rust. This might be a harder issue to resolve than other issues I marked with “good first issue”. Feel free to give it a try though! 

You can open a new PR after you implemented it. Make sure to link back to this issue and the PR I mentioned above.

SirWindfield:
Correction on my part. The Debug implementation doesn’t need to be moved. CliConfig just needs one too that call the one of SharedConfigValues. 

I am typing this up form my memory. So if you think I am talking gibberish, please let me know :)
I just don’t want you people wasting time on an issue that might actually not work as I imagine it right now. That’s why I wrote in #360 that I’ll take a closer look myself :)

Iselk:
Hi @SirWindfield ! Can I help out with this issue?
Should I do as you stated in the first comment?

robinvd:
@Iselk yes that should do it! If you need any help you can ping me here or in a pr

jackvstrickland:
> Hi @SirWindfield ! Can I help out with this issue?
> Should I do as you stated in the first comment?

have you started yet? If not, I wouldn't mind taking a crack at it

SirWindfield:
I'm not actively developing anymore, just keeping a watchful eye around here ;) I've assigned @Iselk as they offered their help first. If we do not receive any updates you can gladly take over @jackvstrickland.

In the meantime, there are other issues that need help :)

Iselk:
Hi @jackvstrickland 

> have you started yet? If not, I wouldn't mind taking a crack at it

I am. It *will* be done this week, *probably* on Friday.

PR #360, opened 2019-10-07
Added option to enable debugging in the login process

TobiMichael96:
This might be a potential fix for #355

I added the option called _debug-login_ (you might want to rename this) which shows the given username, password and password command.

Rewrote some of the code and used a new function for this.

(This is a new pull request because I fucked up some things on the other one... it's possible that this will happen here as well... I'm sorry for that!)

Edit: After I fucked up again, you might need to tell me how to do it properly because at this point it's getting embarrassing...

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -324,9 +324,9 @@ impl FileConfig {
 fn sanitize_user_cred(debug_login: bool, input_value: &Option<String>) -> Option<&str> {
     if input_value.is_some() {
         if debug_login {
-            input_value.as_ref().map(|x| &**x).unwrap();
+            return Some(input_value.as_ref().map(|x| &**x).unwrap());
         } else {
-            "taken out for privacy";
+            return Some("taken out for privacy");
         }
     }
     None


SirWindfield:
Wow, what a PR streak! :)

So I've thought about this for some time today. I feel like the flag should not be a value within the config file and only a CLI flag as it's used for debugging only and not for configuration.

So the ideal case would be to move the `Debug` implementation to `CliConfig`. This is needed as `SharedConfigValues` is part of `CliConfig`. And we could then use your code there. 

I am actually not sure if that would work without problems whatsoever without having the compiler spitting out some errors. So I would rather try this one out at the weekend. That way I am wasting some time and not you if it doesn't work ;)

TobiMichael96:
> Wow, what a PR streak! :)
> 
> So I've thought about this for some time today. I feel like the flag should not be a value within the config file and only a CLI flag as it's used for debugging only and not for configuration.
> 
> So the ideal case would be to move the `Debug` implementation to `CliConfig`. This is needed as `SharedConfigValues` is part of `CliConfig`. And we could then use your code there.
> 
> I am actually not sure if that would work without problems whatsoever without having the compiler spitting out some errors. So I would rather try this one out at the weekend. That way I am wasting some time and not you if it doesn't work ;)

I agree with you, that make a lot more sense. 
I could play around a bit with it and maybe post some snippets in here (to avoid another PR and commit massacre :D), but this might go above my understanding of Rust.

Glad my effort was at least worth something. :)

SirWindfield:
You can create as many commits as you like. I Squash them together before merging anyway :) People always underestimate contributions they make :)

And there is no better way to learn a new language then trying out ;)

TobiMichael96:
I just looked at the code and I think while you are rewriting the `Debug` implementation anyway we should change the line I added in #357.

From `.field("no-audio-cache", &self.no_audio_cache)` to `.field("audio-cache", &!self.no_audio_cache)` in line [349](https://github.com/Spotifyd/spotifyd/blob/master/src/config.rs#L349).

This might clear things up in the output.

SirWindfield:
Thanks for your effort Tobi! I think a move of the Debug might be a better option though. So I’ll close this one for now.

PR #384, opened 2019-10-15
replace device name with default if contains whitespace

### LINKED ISSUE #378: Only allow non-empty device names
I was able to start and use `spotifyd` without a device name, while using user/password. It appears as *blank* in the spotify device list. Can/does this have any negative effects? 

Thanks

juliangaal:
I'm not sure, why the device name can't contain whitespaces. However, I would just replace/ignore device names entered with whitespaces with the default, and warn the user. Which is what this does. Consider this PR highly opinionated and optional :D 

Closes #378.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -511,8 +511,9 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
     let device_name = config
         .shared_config
         .device_name
+        .filter(|s| !s.trim().is_empty())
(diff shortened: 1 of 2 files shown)


SirWindfield:
FYI, [this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b8679ce07755ce5bf0cf4749c9025536) does pretty much the same without introducing a new method :)

SirWindfield:
I've reverted your changes to `utils.rs` and added the above method to filter for whitespaces :)

PR #390, opened 2019-10-16
Run CI only on changes affecting *.rs files

### LINKED ISSUE #380: add path filter to CI workflow to only trigger on *.rs files
https://help.github.com/en/articles/workflow-syntax-for-github-actions#onpushpull_requestpaths

mh84:
Closes #380.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,10 @@
 name: Continuous Integration
 
-on: [pull_request]
-  paths:
-    - '**.rs'
-    
+on:
+  pull_request:
+    paths:
+      - '**.rs'
+
 jobs:
   codestyle:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


SirWindfield:
Thanks for your contribution! would you mind adding the same to the Github Actions workflow files? The relevant documentation is linked here: #380 :)

.github/workflows/ci.yml
```diff
@@ -1,7 +1,9 @@
 name: Continuous Integration
 
 on: [pull_request]
-
+  paths:
+    - '**.rs'
+
```
SirWindfield:
```yaml
on:
  pull_request:
    paths:
    - '**.rs'
```

It probably has to be written this way, as `on: [pull_request]` directly assigns to `on` without allowing more children.

mh84: Alright, that didn't was obvious out of reading the docs.

SirWindfield:
Ye, just noticed that this is actually not completed right as our dependency bot won't trigger new builds now because it only changes toml files. I need to add an exception for those too, ups :) 

Thanks for your help!

PR #400, opened 2019-10-19
Add Rodio Backend

Yarn:


Yarn:
It looks like this actually requires bumping the librespot version. Getting compile errors in librespot with the version in Cargo.lock in master.

SirWindfield:
Note to myself: I should stop working before I actually drank a coffee... Now I had to hard-reset -.-

Yarn:
That version of librespot seems to work fine on osx with rodio at least. I pushed a `cargo.lock` with rodio and it's dependencies. It looks like CI still isn't happy with it still.

SirWindfield:
Thank you for your contribution! The lockfile updated way too much dependencies and I lost track of what was actually happening so I merged the librespot update first and rebased. Kind of messed something up during that 🤔

PR #408, opened 2019-10-21
refactor(utils): refactor get_shell to use native function calls on Linux

SirWindfield:
This commit uses the libc crate to retrieve the users shell instead of relying on parsing the /etc/passwd file


SirWindfield:
Blocked by https://github.com/rust-lang/cargo/issues/1197

PR #409, opened 2019-10-23
README: clearify the space after equal sign

Blacksuan19:
i spent 3 hours woundering what is wrong until i went throw all issues and found this only mentioned in ome of them
something like this should be cleared here in readme


SirWindfield:
This is rather an issue than a feature not mentioned.

PR #418, opened 2019-10-27
Move logging to fern

### LINKED ISSUE #285: Move logging to fern
`fern` is able to log to stdout as well as system logs by attaching handlers during the initialisation. This would move our logging dependency to one crate instead of two and would remove the awkward logging code in `main.rs`. Yes, `fern` does use `syslog` under the hood but it would clean up the `setup_logger` function to only use crate calls from `fern` instead of two different crates.

mh84:
Closes #285

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,9 +47,9 @@ fn setup_logger(log_target: LogTarget, log_level: LevelFilter) {
 fn main() {
     let mut cli_config: CliConfig = CliConfig::from_args();
 
-    let is_deamon = !cli_config.no_daemon;
+    let is_daemon = !cli_config.no_daemon;
 
-    let log_target = if is_deamon {
+    let log_target = if is_daemon {
         LogTarget::Syslog
     } else {
         LogTarget::Terminal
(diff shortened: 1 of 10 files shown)


src/main.rs
```diff
@@ -21,42 +20,55 @@ mod process;
...
+    logger.apply().expect("Couldn't initialize logger");
 }
 
 fn main() {
     let mut cli_config: CliConfig = CliConfig::from_args();
 
-    let is_daemon = !cli_config.no_daemon;
-    let is_verbose = cli_config.verbose;
-    setup_logger(is_daemon, is_verbose);
+    let is_deamon = !cli_config.no_daemon;
```
SirWindfield: It's actually called [daemon](https://en.wikipedia.org/wiki/Daemon_(computing)).

mh84:
Oh sorry, i will fix that typo :)

SirWindfield:
I'll merge it later on this day, after v0.2.20 released. Thanks for your contribution!

SimonPersson:
Thanks, merged!

PR #437, opened 2019-11-25
refactor: report name of selem that cannot be found

pfrenssen:
If a user's volume control is not correctly configured the following (unhelpful) warning is logged:

> Couldn't set volume: "Couldn\'t find selem."

It does not say _which_ selem cannot be found. Let's output this information to help users debug their configuration files.

In my case this PR helped me discover that it was using `PCM` as the name of the selem instead of `alsa`. This led me to #283 which solved my problem.

### DIFF
--- a/src/alsa_mixer.rs
+++ b/src/alsa_mixer.rs
@@ -15,7 +15,7 @@ impl AlsaMixer {
         let mixer = alsa::mixer::Mixer::new(&self.device, false)?;
 
         let selem_id = alsa::mixer::SelemId::new(&*self.mixer, 0);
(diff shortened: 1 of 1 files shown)


pfrenssen:
Build is now green for the newly added code but failing due to a new lint that is added to clippy. This also affects the master branch so I split this off to a separate PR: #438

SirWindfield:
Wow, thanks! I'll try to spare some time off later on in the evening and merge all PRs that are easy to do some. Thanks again for your contribution!

pfrenssen:
Thanks! This should hopefully turn green once #438 is in and merged here.

PR #438, opened 2019-11-25
fix option_and_then_some lint error

pfrenssen:
The latest version of clippy includes the [option_and_then_some](https://rust-lang.github.io/rust-clippy/master/index.html#option_and_then_some) check which is causing recent builds to fail.

Example failing build: https://github.com/Spotifyd/spotifyd/pull/437/checks?check_run_id=319826888#step:6:566

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -495,7 +495,7 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
         .shared_config
         .cache_path
         .map(PathBuf::from)
-        .and_then(|path| Some(Cache::new(path, audio_cache)));
+        .map(|path| Cache::new(path, audio_cache));
 
     let bitrate: LSBitrate = config
         .shared_config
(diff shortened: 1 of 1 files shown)


SirWindfield:
Thanks!

I have no idea what I am missing but my brain keeps telling me that this shouldn't even compile. How does it get transformed into an Option? 🤔

pfrenssen:
I can explain how it works :) Let's take this example:

```
    let pid = config
        .pid
        .and_then(|f| {
            Some(
                f.into_os_string()
                    .into_string()
                    .expect("Failed to convert PID file path to valid Unicode"),
            )
```

`config.pid` is an `Option`. Calling `.and_then()` on it will take any `Some(f)` value from the option, and call the code inside the closure on it (in this case `f.into_os_string()`). We then wrap the result back in `Some()` so it is an `Option` again.

The updated version uses `.map()` which takes the `Some(f)` value, and replaces `f` with the result of the closure. This means the resulting value is still wrapped in `Some()`, we don't need to do this manually!

PR #439, opened 2019-11-27
fix(mpris): position property now returns value in microseconds

backspac:
Position should be returned in microseconds.
See https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:Position

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -495,7 +495,7 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
         .shared_config
         .cache_path
         .map(PathBuf::from)
-        .map(|path| Cache::new(path, audio_cache));
+        .and_then(|path| Some(Cache::new(path, audio_cache)));
 
     let bitrate: LSBitrate = config
         .shared_config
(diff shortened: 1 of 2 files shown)


pfrenssen:
There is already a PR for the clippy warnings, that can be taken out of this PR: #438 

I checked the documentation you linked and can confirm that it specifies the position to be in microseconds. Does this manifest in an actual bug in spotifyd that we can test manually?

SirWindfield:
@pfrenssen It might be that some third-party tools might not detect or work properly with spotifyd if the time returned is not in microseconds. Tools that can control spotifyd over MPRIS (maybe a ncurses based client or something) could show the time passed and returning the wrong time unit would result in them not rendering the passed time properly.

backspac:
@pfrenssen Yes. Both KDEConnect and KDE's Media Player widget were displaying the track position at the beginning of the bar whenever I raised them. playerctl showed the current position as a thousand times smaller.

@SirWindfield Thanks for cleaning up and merging! The MPRIS implementation still has issues. I'm looking into it, but I'm quite new to Rust, so, it might take a while before I understand what's really going on.

SirWindfield:
Great news @backspac. Contributions are always welcome!

Issue #175, opened 2018-10-06
[ERROR] Caught panic with message: connection closed


Trobolit:
After leaving spotifyd running for a few days this message usually shows up along with a crash.
```
[ERROR] Caught panic with message: connection closed
```

I run it on a Ubuntu 16.04.5 LTS server in a VM in proxmox.

To see these crashes I always run it in with this:
```
./spotifyd/target/release/spotifyd --no-daemon --bitrate 320
```

It would be nice if it didn't crash but just reconnected instead, or had the option to shutdown or reconnect on lost connection.

kind regards
Robert

SirWindfield:
This is actually a missing feature of the underlying library `librespot` that is used here. As far as I could tell they still debate on how to make it happen.

NicoHood:
Could you please link the bugreport of the librespot library, so we can watch the fixing progress?

SirWindfield:
If someone in this issue still encounters disconnects that do not reconnect by itself, please re-open this issue!

Joshua-Hwang:
I believe I'm still getting the issued described. After having spotifyd running for about a minute I get the following message
`Caught panic with message: connection closed`

PR #442, opened 2019-12-04
Add proxy support

### LINKED ISSUE #143: Enable http proxy configuration

metiis:
Librespot supports using HTTP proxies to connect to spotify servers. This PR adds a config directive for that. 

This might also help some people who experience an error like in #175  . Helped me at least.

Fixes #143.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2554,6 +2554,7 @@ dependencies = [
  "tokio-core 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-io 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-signal 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 1.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "whoami 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "xdg 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


metiis:
Oh hold up I pushed the wrong version, this one crashes when you don't give it a proxy

metiis:
There, this is it

SirWindfield:
I remember that this didn't work since librespot had an old version of URL that blocked this, see #143. Looks like I missed that they updated the dependency :)

metiis:
Ohh nice thanks, I think #143 was about **https**_proxy though so I didn't mention it as fix

SirWindfield:
I don't thing that librespot works with https 🤔 Or maybe I just misunderstood the old issue haha

Issue #417, opened 2019-10-27
open uri of mpris implementation doesn't work


deisi:
I'm not able to start any song via the mpris interface. I tried:

```
playerctl open https://open.spotify.com/track/225xvV8r1yKMHErSWivnow
playerctl open spotify:track:225xvV8r1yKMHErSWivnow
```

and every combination with and without `"`. is this supposed  to work? spotifyd returns:

```
Couldn't fetch metadata from spotify: Ok(None)
```

username and password for spotify premium are set up. play an pause works. I can start playing via android app and so on. However I need some sort of API or CLI.

deisi:
I have also tried, sp and experience the same. I can pause and play, but I cant start a specific song. However I dont think its a limitation of librespot as ncspot can browse and play songs.

deisi:
by now I think it should work the way I have tried. However it might be another bug due to the refactoring of the mpris implementation. I don't know if it is a regression though, as I have never tested on older versions.

deisi:
@SirWindfield Is this a bug? There is at least a method to open uris:

https://github.com/Spotifyd/spotifyd/blob/d8efb3174a9c0e7ba11cf04f8923ddde96192adf/src/dbus_mpris.rs#L330-L338

SirWindfield:
It probably is, yes. I'll investigate

tommady:
Hi, is this issue still ongoing? since I found the root cause, may I send a PR for this issue?

the root cause is on the device in spotify_api_method macro, it needs the device ID, not the device name.

I plan to add a function to fetch current spotifyd device id by calling the

SirWindfield:
Awesome that you managed to find the cause. Sure, PRs are always welcome.

deisi:
Wohoo that sound awesome :)

Sven Lechner <notifications@github.com> schrieb am Do., 19. Dez. 2019,
10:14:

roqueeee:
Just chiming in to say it would be awesome if someone could resolve this issue.

I'm currently working on a python script that will play a spotify playlist/album/song when an rfid tag encoded with the playlist uri is held to the player.

This bug is pretty much the only thing standing in the way of using spotifyd for the project. For the time being I could use playerctl with the official spotify app, but I'd like to run a headless config with spotifyd in the long run.

Anyways, you guys are doing a great job!

SirWindfield:
Fixed in #459.

PR #459, opened 2019-12-20
fix 417 OpenUri does not work

tommady:
Hi, first of all, I am a newbie of Rust, so if the code is not good enough please guide me 🙇 

this PR is trying to fix this issue
https://github.com/Spotifyd/spotifyd/issues/417

please review, thanks!


tommady:
there has much other lint error, but I don't know should I fix them? since they are not in my modification scope.

VictorKid:
Thanks, man

SimonPersson:
Thanks!

PR #478, opened 2020-01-22
Improve CD workflow matrix (fixes #464)

### LINKED ISSUE #464: spotifyd-linux-full release seems to be missing flags
I noticed that I didn't have the pulseaudio backend available from spotifyd-bin-full from the AUR.

Looking at this CI run: https://github.com/Spotifyd/spotifyd/runs/338348234 

It appears that the same build command is used for both slim and full: `/usr/share/rust/.cargo/bin/cargo +stable build --locked --release --target x86_64-unknown-linux-gnu --no-default-features --features ,alsa_backend`

Building spotifyd-full from source does include the flags.

nicokaiser:
This project uses GitHub Actions to create builds, which is great. However, the workflow for this uses a build matrix that builds every `artifact_type` (slim, full) for every `build_target` (macos, macos-rodio, ubuntu) for every `feature` ('', 'dbus_keyring,dbus_mpris).

So the `cd.yml` creates twice as many workers than necessary, while occasionally overwriting the "slim" version with a "slim" version that has all features enabled (and the other way round), see #464.

This PR improves three things:

* Rename the "ubuntu" target to "linux", so it better represents what it does
* Set the name of the job to "macos-slim", "macos-full", "linux-slim", etc. to give a better overview which jobs are running
* Move the feature definition into the `include` node, this solves a multiplication of workers creating the wrong artifacts

I'll try to adapt the workflow for armhf cross compilation, however this is somewhat complicated because of the system dependencies (alsa, dbus).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -7,18 +7,18 @@ on:
 
 jobs:
   publish:
-    name: Publishing for ${{ matrix.os }}
+    name: Publishing ${{ matrix.build_target }}-${{ matrix.artifact_type }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build_target: [macos, macos-rodio, ubuntu]
+        build_target: [macos, macos-rodio, linux]
         rust: [stable]
-        feature: [
-          '',                                             # Slim version has no features enabled by default.
-          'dbus_keyring,dbus_mpris'                       # Full version has all OS compatible features enabled
-        ]
-        artifact_type: ['slim', 'full']                   # The build strategy will build both types for each OS specified
+        artifact_type: ['slim', 'full']         # The build strategy will build both types for each OS specified
         include:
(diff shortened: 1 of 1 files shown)


SimonPersson:
Thank you!

SirWindfield:
Nice change!

> I'll try to adapt the workflow for armhf cross compilation, however this is somewhat complicated because of the system dependencies (alsa, dbus).

Ye, I tried it a lot on a custom fork so I don't pollute this repository with failed attempts but couldn't get it working. I am not familiar at all with cross compiling toolchains and had a hard time finding up-to-date resources. I could have just copy pasted the Travis configuration to some degree but I didn't want to do that because I didn't really understand why specific flags had to be set.

My attempts revolved around running an ARM build on the workflow servers but that didn't work as I expected it.

It looks like the only way to make this work is to download all dependencies as source files and specify the directories those lie in using compilation flags (probably what the Travis configuration does).

Issue #286, opened 2019-09-09
Provide optimised arm binaries


SirWindfield:
armv7 allows for further optimisation during compilation. We currently only offer armv6 binaries, which run on armv7 too though.

juliangaal:
Which optimizations are these exactly? Getting an armv7 device in the mail today :D

SirWindfield:
That's a hard one I think. At least I couldn't really figure it out. Cross compilation is a bitch sometimes >.>

We currently are on the move to Github Actions (GA) but still use the good old Travis CI. `ci.yml` does some linting and formatting checks for PRs, `cd.yml` does the binary compilation for linux and macos. The ARM binaries are still compiled on Travis CI. 

So this issue would mean to first move the compilation of the ARM binaries to GA. You probably have to create a new workflow as you can't have arm and amd64 binaries installed on the same instance, something like `cd_armv7`. 

> Which optimizations are these exactly?

Ideally you would create three new workflows (armv6, armv7 and aarch64). In theory, armv6 can run on all of them, but each new version introduces new instructions and optimizations the compiler can do. So separate binaries would be ideal. 

Currently, only armv7 gets build on Travis CI.

> But how do I test workflows?

What a great question to ask! You would create a fork of Spotifyd, push a new workflow and create a PR. Make sure to set the trigger to `pull_requests` for testing purposes. Remove the old workflows to shorten run-time. Each new commit to the same PR will trigger a new workflow run. Quite cumbersome yes.

I think the key/trick is to create a workflow for each architecture. I already tried arm compilation but was super greedy and created one workflow for all of them, which didn't work. So I gave up for now :)

Issue #259, opened 2019-06-26
Wrong architecture from v0.2.9 up


nicokaiser:
Release 0.2.5 is the last that runs on the Raspberry Pi 1.
Release 0.2.8 has no binaries.
Release 0.2.9 and up only produce "Segmentation fault" or "Illegal instruction".

I assume something happened to the build pipeline, so the "armv6" binaries are no armv6 binaries anymore. I did not test them on armv7 yet, but will do soon.

Any chances that there will be armv6 binaries for Raspberry Pi 1 / Zero etc.? (If not, it might be a good idea to rename the build artifacts to prevent misunderstandings)

nicokaiser:
Confirmed: the binary works on Raspberry Pi 3, so it is most probably armv7 instead of armv6.

SirWindfield:
Hi all! I can take an hour or two to figure it out completely.
I just noticed though that Github is introducing [Github Actions](https://github.com/features/actions), a pipeline that can be used to run Ci tasks too.
If I get it working without much of a hassle, I'll push a change upstream. But if it takes more time to fix I would propose to wait until we can get access to Github Actions. 

I signed up for the beta already. Not sure how it works with organizations though. @SimonPersson do you have any idea? Did you sign up for it too?`What do you think of my idea to move to Github Acitons?

nicokaiser:
Isn't this was `.travis.yml` does? The "ARM Build" section in this file is supposed to build an ARMv6 build, but I'm not sure if `TARGET=arm-unknown-linux-gnueabihf` really outputs an ARMv6 binary (I have no idea about Rust to be honest).

Maybe the previous builds were ARMv6 compatible by coincidence (they were built for armhf, which is the Debian/Ubuntu platform for ARMv7+VFP3) and just did not use v7-specific commands?

SirWindfield:
The target is certainly the right one. I did some research and it looks like it depends on the toolchain used to cross compile. Ubuntu's `armhf` doesn't only implement hard-floating specific instruction sets for ARMv7, it adds some instructions outside of the `hf` scope too, which makes it incompatible with ARMv6 chips that supported the `hf` instruction set. I remember seeing a toolset that allowed to compile for ARMv6 specifically...

opsydev:
Hey,

Is there any update on this?

rprofijt:
Hi, it seems that after v0.2.19 there are no arm builds anymore, is that related?

nicokaiser:
[...]

Issue #244, opened 2019-05-01
Dbus not avalible when headless


insoPL:
Session DBUS is not avalible without X11. It would be neat to add option in config to make MPRIS avalible through system dbus. When I edited dbus_mpris.rs 
`    let c = Rc::new(Connection::get_private(BusType::Session).unwrap());
`
and changed Session to System and recompiled it everything worked great. I would make pull request but i cannot into rust so it will be mess anyway. Please can somone make it? It's big deal for headless raspberry users.

ius:
I just ran into this as well (due to Arch/AUR enabling all features):

````
Jun 14 13:09:42 system spotifyd[11995]: 13:09:42 [ERROR] Caught panic with message: called `Result::unwrap()` on an `Err` value: D-Bus error: Server address of type unix was missing argument path or abstract (org.freedesktop.DBus.Error.BadAddress)
Jun 14 13:09:42 system systemd[1]: spotifyd.service: Main process exited, code=exited, status=101/n/a
````

Interestingly it still doesn't work for me if a user DBUS session is available (ie. when run as root via SSH) - it will start and become visible in proprietary Spotify clients, but exits as soon as playback is switched to spotifyd.

I don't see an obvious error in the debug logs, but the *Ok(NotReady)* does seem suspicous. When I rebuild *without* dbus_mpris and dbus_keyring everything works fine.

````
12:56:37 [DEBUG] dbus_tokio::adriver: Polling message stream
12:56:37 [DEBUG] dbus_tokio::adriver: msgstream found Ok(NotReady)
12:56:37 [DEBUG] tokio_reactor::registration: scheduling Write for: 6
12:56:37 [TRACE] mio::poll: [<unknown>:905] deregistering handle with poller
12:56:37 [DEBUG] tokio_reactor: dropping I/O source: 6
12:56:37 [DEBUG] librespot_connect::spirc: drop Spirc[0]
12:56:37 [DEBUG] librespot_playback::player: Shutting down player thread ...
12:56:37 [DEBUG] librespot_playback::player: drop Player[0]
12:56:37 [DEBUG] librespot_core::session: drop Session[0]
12:56:37 [DEBUG] librespot::component: drop MercuryManager
12:56:37 [DEBUG] dbus_tokio::adriver: Dropping AMessageStream
12:56:37 [DEBUG] dbus_tokio::adriver: AMessageStream telling ADriver to quit
[...]

PR #479, opened 2020-01-22
add CD support for armv6 and armv7

nicokaiser:
This PR adds a workflow that creates "arm-unknown-linux-gnueabihf" builds for the "slim" and "full" versions with "alsa_backend" using cross compilation.

The resulting packages are called `spotifyd-linux-armhf-slim.tar.gz` and `spotifyd-linux-armhf-full.tar.gz`.

As a new package source has to be installed, `ubuntu-latest` cannot be used here, so the new step uses `ubuntu-18.04` (which currently is the same, but we need to be sure).

Instead of a full `sbuild` setup, the armhf packages are installed and copied manually, which may look dirty. Unfortunately, `dbus-rs` needs quite some system packages, so we need to download and put them where the linker expects them to be. This process is described here: https://github.com/diwic/dbus-rs/issues/184#issuecomment-520228758

Issues:

- I'm not sure if the binaries produced by this run on the Raspberry Pi 1 (or Zero). To be able to run on these, it might be necessary to use the [arm-bcm2708](https://github.com/raspberrypi/tools/tree/master/arm-bcm2708) linker from raspberrypi/tools, as stated here: https://github.com/BurntSushi/ripgrep/issues/676#issuecomment-374058198
- Can someone test the `full` version? I can build it, but cannot get MPRIS D-Bus running on my machine, which may as well be a client problem: the process refuses to start without `$DISPLAY`, and when using the system dbus, `playerctl` acts up and won't display anything. Not sure if I broke this or it did not work at all in the first place. (If I understand it correctly, starting the `full` version *always* requires D-Bus? #244)

Related: #259, #280, #286

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -56,36 +56,36 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y -qq libasound2-dev libssl-dev libpulse-dev libdbus-1-dev
       - name: Installing needed Ubuntu armhf dependencies (slim)
-        if: matrix.os == 'ubuntu-18.04' && matrix.target == 'arm-unknown-linux-gnueabihf'
+        if: matrix.os == 'ubuntu-18.04' && matrix.build_target == 'linux-armhf'
         run: |
-          sudo mkdir -p /build/deps_root
+          sudo mkdir -p /build/sysroot
           echo "deb [arch=armhf] http://ports.ubuntu.com/ubuntu-ports/ bionic main" | sudo tee -a /etc/apt/sources.list
           sudo apt-get update
           sudo apt-get install -y -qq gcc-arm-linux-gnueabihf libc6-armhf-cross libc6-dev-armhf-cross
           sudo apt-get download libasound2:armhf libasound2-dev:armhf libssl-dev:armhf libssl1.1:armhf
-          sudo dpkg -x libasound2_*.deb /build/deps_root/
-          sudo dpkg -x libssl-dev*.deb /build/deps_root/
-          sudo dpkg -x libssl1.1*.deb /build/deps_root/
-          sudo dpkg -x libasound2-dev*.deb /build/deps_root/
+          sudo dpkg -x libasound2_*.deb /build/sysroot/
+          sudo dpkg -x libssl-dev*.deb /build/sysroot/
+          sudo dpkg -x libssl1.1*.deb /build/sysroot/
+          sudo dpkg -x libasound2-dev*.deb /build/sysroot/
       - name: Installing needed Ubuntu armhf dependencies (full)
-        if: matrix.os == 'ubuntu-18.04' && matrix.target == 'arm-unknown-linux-gnueabihf' && matrix.artifact_type == 'full'
+        if: matrix.os == 'ubuntu-18.04' && matrix.build_target == 'linux-armhf' && matrix.artifact_type == 'full'
         run: |
           # Make dbus-rs cross-compile, see https://github.com/diwic/dbus-rs/issues/184#issuecomment-520228758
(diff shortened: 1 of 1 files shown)


nicokaiser:
The Travis workflow was deactivated anyway, right? So if (?) this works, the `.travis.yml` could also be removed.

.github/workflows/cd.yml
```diff
@@ -34,40 +34,87 @@ jobs:
             artifact_prefix: linux
             audio_backend: alsa
             target: x86_64-unknown-linux-gnu
+          - build_target: linux-armhf
+            os: ubuntu-18.04
```
nicokaiser: Since "bionic" is needed when adding the Ubuntu Ports source later, this must be hard-coded to `ubuntu-18.04`.

.github/workflows/cd.yml
```diff
@@ -34,40 +34,87 @@ jobs:
...
+          RUSTFLAGS: "-C linker=arm-linux-gnueabihf-gcc -L/usr/arm-linux-gnueabihf/lib -L/build/deps_root/usr/lib/arm-linux-gnueabihf -L/build/deps_root/lib/arm-linux-gnueabihf"
+          PKG_CONFIG_ALLOW_CROSS: 1
+          OPENSSL_LIB_DIR: /build/deps_root/usr/lib/arm-linux-gnueabihf
+          OPENSSL_INCLUDE_DIR: /build/deps_root/usr/include/arm-linux-gnueabihf
       - name: Packaging final binary
         shell: bash
         run: |
           cd target/${{ matrix.target }}/release
-          
-          strip spotifyd
```
nicokaiser: `strip` will not recognise cross-compiled binaries, so we cannot always use it (and just assume that Rust creates good-enough ones).

SirWindfield:
Great work! Yes, the current DBUS doesn't work in headless for most cases. I've tried the same with `ubuntu-16.04` but I didn't get it to work :)



>     * I'm not sure if the binaries produced by this run on the Raspberry Pi 1 (or Zero). To be able to run on these, it might be necessary to use the [arm-bcm2708](https://github.com/raspberrypi/tools/tree/master/arm-bcm2708) [](/raspberrypi/tools/tree/HEAD@{2020-01-22T18:00:49Z}/arm-bcm2708)

I think I've posted this somewhere in one of our issues so I don't forget this. That is the case, yes. The binary you created should be armv7 iirc. If you have the time, could you try to run it in a qemu armv6 emulator and see if it loads? Because I am actually not sure what target armhf exactly targets.

THis could be expanded to support aarch64 as well as most dependencies should be available there as well. The tricky part is to find a working mirror that provides those files for us. That was the main issue I faced with my try.

SirWindfield:
Would it be possible to extend this for armv6 as well @nicokaiser?
Relevant (maybe): https://github.com/Spotifyd/spotifyd/issues/259#issuecomment-531956408

SirWindfield:
@nicokaiser could you trigger a new release on your fork and see if the workflow works correctly?

nicokaiser:
Okay the `linux-armhf` version does not work on the Raspberry Pi Zero (and thus, neither on the Pi 1).

I managed to add a build for armv6 using the Raspberry Pi toolchain, however due to the very old gcc version (gcc-4.9.3 from 2015) and its libc-2.19 the `full` version will not compile, as systemd requires a newer libc.

There are other repositories with updated toolchains (such as [Pro/raspi-toolchain](https://github.com/Pro/raspi-toolchain)), however I think the `slim` version will be enough for the most usecases of spotifyd on a Raspberry Pi 1/Zero, so I'll just exclude the `full` version for the `linux-armv6` builds.

SirWindfield:
Awesome! Thank you very much!

SirWindfield:
I forked your repo and ran a test. All binaries sucessfully built and were deployed. 

Thank you very very much! You did an awesome job here! Really appreciated!

Issue #450, opened 2019-12-13
add support for device-type


wdehoog:
Librespot has the option to specify the device type:

```
The displayed device type in Spotify clients. Can be unknown, computer, tablet, smartphone, speaker, tv, avr (Audio/Video Receiver), stb (Set-Top Box), and audiodongle. Defaults to speaker 
```

I am running spotifyd on my UBports phone and it would be nice to be able to configure it to `smartphone`.

If this is not yet possible how could it be added?

SirWindfield:
You'd have to add a new configuration to the `SharedConfigValues` struct:
https://github.com/Spotifyd/spotifyd/blob/a789b49c3d8bd0e4efdbb41587e5f4f61f602ff4/src/config.rs#L207-L208.

Additionally, all possible values would have to be listed inside of an `enum` (for example called `DeviceType`) similar to the `backend` and `bitrate` options:
https://github.com/Spotifyd/spotifyd/blob/a789b49c3d8bd0e4efdbb41587e5f4f61f602ff4/src/config.rs#L120-L151

The important part is to implement `From<DeviceType> for librespot::core::config::DeviceType` as that allows to convert the values without much of an effort when passing down to `librespot`.

You'd then have to add a new field to the `SpotifydConfig` struct that holds the device type. Inside of `setup.rs` the main loop state is created. The value has to be set in there as well using a new struct field. The last step would be to pass down that value inside of `main_loop.rs`:
https://github.com/Spotifyd/spotifyd/blob/a789b49c3d8bd0e4efdbb41587e5f4f61f602ff4/src/main_loop.rs#L166-L172

wdehoog:
Hi. I have done some work already. Please see https://github.com/wdehoog/spotifyd/tree/add_device-type 
But I cannot build the armhf version I'd like to test on my phone so I did not make a PR yet.

SirWindfield:
Ah, fair enough! Let me know if you make any progress!

PR #496, opened 2020-02-05
add documentation for running on macOS

johnalanwoods:
I have created docs to help new users run as a daemon on macOS (which is the whole point of spotifyd!

It's a little tricky, so tried to call out gotchas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Spotifyd streams music just like the official client, but is more lightweight an
 - [Configuration](#configuration)
   - [CLI options](#cli-options)
   - [Configuration file](#configuration-file)
-- [Running as a systemd service](#running-as-a-systemd-service)
+- [Running as a system service](#running-as-a-system-service)
 - [Common issues](#common-issues)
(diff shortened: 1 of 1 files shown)


SirWindfield:
Looks good, thank you! Some notes for later on when we do the configuration flag renaming:

1) Stuff like this should probably go into the Wiki instead of the readme. The readme is quite large by now.
2) Example configuration files for each OS

johnalanwoods:
Thanks @SirWindfield 👍 

Points noted... Hope to contribute further, cool product.

SirWindfield:
The notes are just a reminder for myself, don't worry! There isn't really much going on in the Wiki right now but I hope we can move a lot of the configuration specific stuff into it (like compiling on specific platforms or special cases like this one :smile:).

Contributions are always appreciated. Currently there aren't a lot of `good-first-issue`s around as almost all of them require some more thought put into them from my side. That being said, https://github.com/Spotifyd/spotifyd/issues/450 seems like a good candidate ;)

PR #500, opened 2020-02-06
feat: add support for device-type

wdehoog:
See Issue #450. 

Sorry for all the git noise. I am unfit for git. For rust as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,11 @@ zeroconf_port = 1234
 
 # The proxy `spotifyd` will use to connect to spotify.
 proxy = http://proxy.example.org:8080
+
+# The displayed device type in Spotify clients.
+# Can be unknown, computer, tablet, smartphone, speaker, tv,
+# avr (Audio/Video Receiver), stb (Set-Top Box), and audiodongle.
+device_type = speaker
 ```
 
 #### Alternatives to storing your password in the config file <!-- omit in toc -->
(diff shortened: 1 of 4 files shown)


SirWindfield:
The whole git noise is no problem. I usually squash all commits into one (merging them) before merging into master. Better to split them off into smaller ones that are easier to follow then having one huge commit. I would still consider this commit small so don't worry :)

Could you run `cargo fmt` as well as `cargo clippy`? The first formats the code and the second command runs some linting. Usually you get some warnings displayed in your terminal with suggestions on how to fix them. Quite straight forward.

Would be nice if you could do that so the CI passes :smile:

I wouldn't consider myself good writing Rust code either. There are a lot of people who are way better at it than me. But you have to start somewhere and open source is a great place to gather experience for languages you are not that familiar with :)

SirWindfield:
They should both be installed if you use the stable branch. If not, `rustup component add clippy rustfmt` should do the trick.

wdehoog:
Thanks. I'll run these tools. I tried to keep the formatting the same as already present. This fmt tool seems to disagree. Do you prefer the way fmt advises?

SirWindfield:
Yep, we always run `cargo fmt`. Took a glimpse it seems that the only problem was that your imports were not alphabetically sorted, which should `cargo fmt` do for you :)

In the feature this will be done automatically (#501).

SirWindfield:
Thanks!

PR #525, opened 2020-02-28
feat: add client_id environment variable

### LINKED ISSUE #507: add a way to let users use their own Spotify client id for API requests
Currently one single token for all users is used and the rate limit can be reached that way. Users should be able to use their own client id.

Should the current one be removed completely?

SirWindfield:
Users can now pass down their own client id to circumvent rate limit errors

fix #507

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -25,7 +25,7 @@ use rspotify::spotify::{
 };
 use tokio_core::reactor::Handle;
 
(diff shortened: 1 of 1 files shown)


yousefamar:
I'm guessing the original idea was that this is just temporary until it's been added to as part of spotifyd.conf or something, but it seems to have been around for a while and it took me a while to find out this was an option. Should I PR to document this in the readme?

SirWindfield:
Yep, definitely :)

PR #553, opened 2020-03-30
feat: username_cmd config option

maringuu:
The code is copy pasted from password_cmd.
In the first commit the doubled check for an empty `password_cmd` is removed.


SirWindfield:
Thanks for your contribution and sorry that it took that long to merge!

Issue #402, opened 2019-10-20
Move configuration file format to TOML


SirWindfield:

slondr:
Included in v0.2.25

PR #571, opened 2020-04-16
TOML config

0jdxt:
I had a problem with '#' being in my password and i saw it was an issue with serde_ini and in [this](https://github.com/Spotifyd/spotifyd/issues/402) milestone so did this. i think it works

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ libc = "0.2.73"
 log = "0.4.6"
 percent-encoding = "2.1.0"
 rspotify = "0.8.0"
-serde = { version = "1.0.103", features = ["derive"] }
-sha-1 = "0.8"
-structopt = "0.3.11"
+serde = { version = "1.0.115", features = ["derive"] }
+sha-1 = "0.9"
+structopt = "0.3.17"
 syslog = "4"
 tokio-core = "0.1"
 tokio-io = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ log = "0.4.6"
 percent-encoding = "2.1.0"
 rspotify = "0.8.0"
 serde = { version = "1.0.103", features = ["derive"] }
-serde_ini = "0.2.0"
 sha-1 = "0.8"
 structopt = "0.3.11"
 syslog = "4"
@@ -35,6 +34,8 @@ tokio-signal = "0.1"
 url = "1.7"
 xdg = "2.2"
 librespot = { version = "0.1.1", default-features = false, features = ["with-tremor"] }
+toml = "0.5.6"
+serde_repr = "0.1.5"
 
 [target."cfg(target_os = \"macos\")".dependencies]
 whoami = "0.8.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ url = "1.7"
 xdg = "2.2"
 librespot = { version = "0.1.1", default-features = false, features = ["with-tremor"] }
 toml = "0.5.6"
-serde_repr = "0.1.5"
 
 [target."cfg(target_os = \"macos\")".dependencies]
 whoami = "0.9.0"
--- a/README.md
+++ b/README.md
@@ -169,19 +169,19 @@ spotifyd --help
 
 ### Configuration file
 
-`Spotifyd` is able to load configuration values from a file too. The file has to be named `spotifyd.conf` and reside in the user's configuration directory (`~/.config/spotifyd`) or the system configuration directory (`/etc` or `/etc/xdg/spotifyd`). This also applies to macOS!
+`Spotifyd` is able to load configuration values from a TOML file too. The file has to be named `spotifyd.conf` and reside in the user's configuration directory (`~/.config/spotifyd`) or the system configuration directory (`/etc` or `/etc/xdg/spotifyd`). This also applies to macOS!
 
 The configuration file consists of two sections, `global` and `spotifyd`, whereas `spotifyd` takes priority over `global`.
 
 The configuration file has the following format:
 
-```ini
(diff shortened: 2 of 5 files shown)


0jdxt:
sorry I used the github app and the 'close' button hella confused me

SirWindfield:
I'll merge this into a WIP branch for the major release this week! Not sure if `serde_repr` does change stuff internally. It seems like it has the same effect as renaming (as was done before).

0jdxt:
> I'll merge this into a WIP branch for the major release this week! Not sure if `serde_repr` does change stuff internally. It seems like it has the same effect as renaming (as was done befo( re).

yay! I used serde_repr so both structopt and toml take an integer, otherwise toml would expect a string. But maybe theres another way, for me changing the renames to repr worked.

src/config.rs
```diff
@@ -480,44 +455,24 @@ impl fmt::Debug for SharedConfigValues {
...
-                l.drain(..last_index).collect()
-            })
-            // The password field takes the whole value as the password. We need to remove the space between
-            // the password and the # character.
-            .map(|l: String| l.trim().to_string())
-            .collect();
-
-        let comment_free_content = comment_free_lines.join("\n");
-        let config_content: FileConfig = serde_ini::from_str(&comment_free_content).unwrap();
+        let config_content: FileConfig = toml::from_str(&content).unwrap();
```
Stupremee, line 475: I don't think it should panic here. 
Reporting the error would make much more sense

Stupremee (changes requested):

src/config.rs
0jdxt: Yeah it would be ideal for the toml errors to be propagated to the user and inform them how to fix the file.

0jdxt: Feel free to request a PR on my fork

SirWindfield: I'd probably use ˋexceptˋ instead of unwrap and write something like "error during deserialization". The except call does propagate the error message as well do the user would have an idea what is wrong :)

Cargo.toml
```diff
@@ -24,17 +24,18 @@ libc = "0.2.73"
 log = "0.4.6"
 percent-encoding = "2.1.0"
 rspotify = "0.8.0"
-serde = { version = "1.0.115", features = ["derive"] }
-serde_ini = "0.2.0"
-sha-1 = "0.9"
-structopt = "0.3.17"
+serde = { version = "1.0.103", features = ["derive"] }
+sha-1 = "0.8"
+structopt = "0.3.11"
```
Plecra, line 29: These dependencies need to be brought in line with master

Cargo.toml
```diff
@@ -24,17 +24,18 @@ libc = "0.2.73"
...
 syslog = "4"
 tokio-core = "0.1"
 tokio-io = "0.1"
 tokio-signal = "0.1"
 url = "1.7"
 xdg = "2.2"
 librespot = { version = "0.1.1", default-features = false, features = ["with-tremor"] }
+toml = "0.5.6"
+serde_repr = "0.1.5"
```
Plecra, line 38: In this case, I'd prefer a manual `Deserialize` implementation. It's easier for a dev to understand

```rust
impl<'de> Deserialize<'de> for Bitrate {
    fn deserialize<D: de::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        match u16::deserialize(deserializer) {
            96 => Ok(Self::Bitrate96),
            160 => Ok(Self::Bitrate160),
            320 => Ok(Self::Bitrate320),
            unknown => Err(D::Error::invalid_value(
                de::Unexepected::Unsigned(unknown.try_into().unwrap()),
                "a bitrate"
            )),
        }
    }
}

File: src/config.rs (hunk 1 above)
Comment (Plecra) [reply src/config.rs:475]: I'm happy to leave this as-is for now. Instead, it'd be good to make a proper error reporting system that can be more user-friendly than panics

**Review by Plecra — CHANGES_REQUESTED** [review]
I like it 😀 Are there working configs that will work differently after this PR?

**Comment by 0jdxt** [comment]
> I like it  Are there working configs that will work differently after this PR?
I would imagine not since ini is a subset of toml but any non-number value must be in quotes whereas it doesn't matter for ini.

File: Cargo.toml (hunk 3 above)
Comment (0jdxt) [reply Cargo.toml]: done

**Comment by robinvd** [comment]
@0jdxt the readme now has this line
```
username = username
```
should that then become
```
username = "username"
```

**Review by robinvd — CHANGES_REQUESTED** [review]
@0jdxt the readme now has this line
```
username = username
```
should that then become
```
username = "username"
```

**Review by robinvd — APPROVED** [review]
Looks good! So just for clarity, this will be a breaking change.

**Comment by 0jdxt** [comment]
indeed

**Review by robinvd — APPROVED** [review]

**Review by slondr — APPROVED** [review]

File: README.md [hunk 4]
```diff
@@ -169,27 +169,27 @@ spotifyd --help
 
 ### Configuration file
 
-`Spotifyd` is able to load configuration values from a file too. The file has to be named `spotifyd.conf` and reside in the user's configuration directory (`~/.config/spotifyd`) or the system configuration directory (`/etc` or `/etc/xdg/spotifyd`). This also applies to macOS!
+`Spotifyd` is able to load configuration values from a [TOML](https://toml.io/en/v0.5.0) file too. The file has to be named `spotifyd.conf` and reside in the user's configuration directory (`~/.config/spotifyd`) or the system configuration directory (`/etc` or `/etc/xdg/spotifyd`). This also applies to macOS!
```
Plecra, line 172: Now we've made breaking changes, I'd like to use the ``.toml` extension. That doesn't need to be part of this PR though

Plecra (approved):
Thanks for the PR! This is a great first step for better UX

Issue #457, opened 2019-12-17
DBus PropertiesChanged event not fired


mainrs:
**Description**

Metadata changes do not raise the `org.freedesktop.DBus.Properties.PropertiesChanged` event

**Compilation flags** 
- [x] dbus_mpris
- [ ] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [ ] pulseaudio_backend
- [ ] rodio_backend

**Versions (please complete the following information):**
 - OS: 
 - Spotifyd: 
- cargo: 

See https://github.com/Spotifyd/spotifyd/issues/455

justin-gerhardt:
I've tried implementing this a bit but I'm over my head with rust.
The dbus library has native support for the propertiesChanged event.
The properties have to have tagged `.emits_changed(dbus::tree::EmitsChangedSignal::True) `(invalidate doesn't work with playerctl)

Normally the signal is sent by the library automatically on property modification from dbus. In this case since the change is internal it has to be manually sent.
A relevant github issue https://github.com/diwic/dbus-rs/issues/66
[The only result on github I could find that uses the API](https://github.com/qzed/linux-surface-dtx-daemon/blob/d8e67ffff5f2ff98f6860f09c6f2a3bd8a1d75ce/surface-dtx-daemon/src/service.rs#L86)

Ideally the signal would be raised for the metadata, playback status and position properties.

SirWindfield:
Would you mind trying out if the version over at https://github.com/Spotifyd/spotifyd/tree/fix_propertieschanged_event ? This might work but I can't really try it because I don't have access to a linux env right now @justin-gerhardt

justin-gerhardt:
Setting emits_changed alone only raises the signal if the property change was caused by a dbus message setting the metadata property (which can't happen, it's read only). dbus-rs has no way of knowing that fetching the underlying metadata would result in a new result.

You will have to somehow feed metadata change events into the dbus server and then run something along the lines of 

```
let mut chg = Vec::new();
property_metadata.clone().add_propertieschanged(&mut chg, media_player2_player_interface.get_name(), ||{
     Box::new(// new metadata)
});
let msg = chg.first().unwrap().to_emit_message(object_path.clone().get_name());
connection.clone().send(msg).unwrap();
```

JasonLG1979:
> Ideally the signal would be raised for the metadata, playback status and position properties.

DO NOT emit a property changed signal for position. That would result in a crazy amount of signals. 

[...]

PR #575, opened 2020-04-20
WIP: support PropertiesChanged event

bbigras:
Here's maybe a way #457 could be done.

The code is pretty ugly right now but maybe the devs can help improve it or base the real work on this.

I think I shouldn't use `continue` since the commented code wouldn't run otherwise.

Right now I only tested getting the metadata when I skip to the next track.

And right now I'm more interested to know if it's the right way or not before trying to make it better.


bbigras:
I would like to spend some more time on this. Any comments from the devs if I'm on the right track?

I guess that if I make a fully functional PR, it could always be adapted to the correct way later.

backspac:
Nice!
I'm not a dev but I'd take a look at [ncspot's implementation](https://github.com/hrkfdn/ncspot/blob/master/src/mpris.rs). It might give you ideas and it seems quite clean.
Also, I'd build it against the [dev branch of librespot](https://github.com/librespot-org/librespot/tree/dev) or [this one](https://github.com/ashthespy/librespot/tree/fix_protobuf) (for the protobuf fix) as they have implemented new events and we could be about to [bump librespot versions](https://github.com/Spotifyd/spotifyd/pull/596).

Maik1999:
I build the branch and got the problem that GNOME does not display the artist.
I'm not familiar with rust so I can just guess but as far as I can tell the artists are not send as a string list but as a simple string. See the [mpris spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index6h4) for more information.

bbigras:
Thanks for the interest @Maik1999. I'll take a look.

SirWindfield:
Sorry for not responding @bbigras. I totally missed this PR :(

bbigras:
> I build the branch and got the problem that GNOME does not display the artist.
> I'm not familiar with rust so I can just guess but as far as I can tell the artists are not send as a string list but as a simple string. See the [mpris spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index6h4) for more information.

@Maik1999: Are you using a gnome extension? I don't remember if gnome supports mpris natively.

Do you have an example of a track no showing the artist's name?

It seems to work for me:

image



Not sure which extension and version I'm using but it looks like this in gnome-tweaks:

image

bbigras:
> Sorry for not responding @bbigras. I totally missed to answer :

@SirWindfield: Hey no worries. :smile: Thanks for replying now.

I'm interested in knowing if I'm on the right track or not.

After I started to work on this PR I also saw that spotifyd had a `--on-song-change-hook` option. I was wondering I should use the same code.

Also I think my code only works when the track changes, not when it starts playing.

And I need to add the other stuff: thumbnail, album...

Maik1999:
I have a notification in the notification menu but I am not sure if it is directly GNOME or if Manjaro has installed something by default. Tried the MPRIS indicator button and there it works with the artist name.

bbigras:
> I have a notification in the notification menu but I am not sure if it is directly GNOME or if Manjaro has installed something by default. Tried the MPRIS indicator button and there it works with the artist name.

If you test with the web version of Spotify, do you see the artist's name in the notification in the menu?

Maik1999:
> > I have a notification in the notification menu but I am not sure if it is directly GNOME or if Manjaro has installed something by default. Tried the MPRIS indicator button and there it works with the artist name.
> 
> If you test with the web version of Spotify, do you see the artist's name in the notification in the menu?

Actually there is no notification at all with Firefox but the official Spotify client shows the arists name.

izissise:
I tested this changes and it works as intended, do you need help on something in order to merge this?

bbigras:
@izissise Are you part of the new team? In https://github.com/Spotifyd/spotifyd/pull/575#issuecomment-658354365 I would have liked to have some feedback about if I was on the right track and if `--on-song-change-hook` was fully functional and could be used for the same thing.

izissise:
I'm not part of the new team, however I'll try to give some feedback :)

bbigras:
Here's an example of using `--on-song-change-hook`. I'll try to find some time for this PR soon.

```
Running "/home/bbigras/spot.sh" using "/run/current-system/sw/bin/fish" with environment variables {"OLD_TRACK_ID": "0L7TjIdcVQaDlX3nQbMdQC", "PLAYER_EVENT": "change", "TRACK_ID": "73Nyo2ajASwg4WFHnzVKFo"}
Running "/home/bbigras/spot.sh" using "/run/current-system/sw/bin/fish" with environment variables {"PLAYER_EVENT": "stop", "TRACK_ID": "73Nyo2ajASwg4WFHnzVKFo"}
```

izissise:
I can confirm that the hook is correctly run with the env vars when changing the song through the dbus interface

src/dbus_mpris.rs
```diff
@@ -610,6 +637,88 @@ fn create_dbus_server(
...
+        match message {
+            TrackChange::Changed(changed_track) => {
+                println!("TrackChange::Changed");
+
+                let track_name2: Variant<Box<dyn RefArg>> =
+                    Variant(Box::new(changed_track.track.to_string()));
+
+                let album: Variant<Box<dyn RefArg>> =
+                    Variant(Box::new(changed_track.album.to_string()));
+                let artist: Variant<Box<dyn RefArg>> = Variant(Box::new(changed_track.artist));
```
scaryrawr, line 652: For the GNOME media control, this only works for the initial song playing for me, for it to work on track changes, I modified it to:

```
let artist: Variant<Box<dyn RefArg>> = Variant(Box::new(vec![changed_track.artist.to_string()]));
```

It looks like in the artist should be a [list of strings](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#index6h4).

(I don't know rust, but really want this feature!).

izissise:
Still needed

izissise:
This still needs to be solved, for proper MPris Support

src/main_loop.rs
```diff
@@ -24,6 +24,9 @@ use std::{io, rc::Rc};
 use tokio_core::reactor::Handle;
 use tokio_io::IoStream;
 
+#[cfg(feature = "dbus_mpris")]
+use crate::dbus_mpris::{ChangedTrack, DbusServer, TrackChange};
```
shreyasminocha, line 28: I tried building this with `cargo build --release --all-features` (rustc 1.53.0), but it complained:

```
error[E0252]: the name `DbusServer` is defined multiple times
  --> src/main_loop.rs:28:39
```

…and

```
warning: unused import: `DbusServer`
  --> src/main_loop.rs:28:39
```

Am I missing something?

evelez7:
Update: After poking around, I've found that there is a built-in struct in [`dbus-rs`](https://github.com/diwic/dbus-rs) which would greatly simplify the implementation of notifying about changed properties. The problem is that it is on a version greater than what is currently used in spotifyd. Compiling with newer versions breaks the current mpris code rather badly.

I'm not sure if this is something that the maintainers see as outside of the scope of the project, but it seems a necessary change for reasonable integration with MPRIS. This PR's implementation is unnecessarily complicated with what can be provided.

PR #600, opened 2020-06-09
spotifyd.service as an user service

awerlang:
As an user service, it can't depend on the system service `network-online.service`. Also, fail the service right away if it's not the correct path. And assume a better place for the binary (distro packages can change this anyway).

This is intended to run as an user service, another .service can be proposed for a system-wide service, with sandboxing settings.


SirWindfield:
Thanks for this PR first of all! I wanted to chime in and ask for some stuff and see what your view point is:

I do agree that `spotifyd` shouldn't probably run as a system service. The current `service` file works great for most of the people without having to specify extra custom service files that mimick system services like `network-online` or `audio`.

What do you think of the proposal of moving the current file into a subdirectory like (e.g. `systemd/system`) and create a new one (the one you have in this PR) under `systemd/user`. 

This would allow more advanced people or people that want to have it as a user service to copy that file and people that want it as a system service copy the other one.

This change would need to cover as well:
- [ ] Documentation changes in the readme (especially commands would need changes as they refer to the current path)
- [ ] Mention the existence of both user and system services
- [ ] Briefly (!) mention the difference between the two and maybe link to an external resource to read up on this as well. A sentence stating which version should be preferred if you want a working out-of-the-box experience should be added as well (which I guess would be system).

What do you think?

awerlang:
Fair points, I updated the PR addressing your requests.

I removed the part mentioning (distro) package maintainers. From these docs they'll know what correct paths are, which shall be under `/usr/lib/systemd/`, but would only add noise here.

And in my experience, the user service is easier to get it working. I manage to have it working as a system service, but it was rather cumbersome since pulseaudio is an user service here. The mileage may vary for each distro. Hopefully someone which managed to run the service system-wide will contribute back a working recipe, be it with alsa or pulseaudio.

PR #607, opened 2020-06-11
[WIP]: fix shell script lints

SirWindfield:
This fixes some `spellcheck` lints. I can't test linux/posix related stuff right now due to a bug inside of WSL and I don't really want to merge untested stuff (even though this is a simple change). But I never used this `sh` syntax so I am actually not 100% sure that it works as expected.

If someone could try the script or (honestly) confirm that this works indeed as it should (take a look at the file diffs), then I'd be willing to merge it.

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -24,8 +24,7 @@ echo "→ Checking for local changes..."
 require_clean_work_tree
 
 echo "→ Formatting Rust code..."
-cargo fmt
-if [ $? -ne 0 ]; then
+if ! cargo fmt; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


carlsmedstad:
I have confirmed this works on Linux.

A, in my mind, further improvement would be to use some or all of:
```sh
set -eux
```

* `e` would remove the need for all if-statment + `exit 1` since it will error if any command produces a non-zero exit code.
* `u` to produce an error should any variable not be undefined, mostly for debug purposes.
* `x` to print all commands to stdout. This greatly reduces the need for echoing what is being done, but might be to verbose to some tastes.

Issue #445, opened 2019-12-06
Initial volume always at 100%


mzanetti:
When starting spotifyd the initial volume setting is always 100%. This might be *very* loud, depending on the speakers.

Now, in addition to that, spotifyd seems to crash rather often and is respawned by systemd, which causes volume to reset itself to 100%.

Shouldn't the last used volume be cached somehow and restored on startup?

SirWindfield:
It should be restored, yes. IIRC this is handled inside of `librespot` as well. Do you have crash logs available? `systemd` should contain some logging output. Make sure to start the daemon using `RUST_BACKTRACE=1 spotifyd --verbose` so I get additional output :)

uiii:
The same is happening to me. E.g. when I start the RPi.

mzanetti:
Ok, Sure, I can report the crashes too. But this actually bug is not about crashes. It also happens on restarts, as @uiii reported too. Also if I just ctrl+c spotifyd and start it again. Volume is always at 100%.

mzanetti:
> It should be restored, yes. 

Ok, so that's a bug then, it clearly is not.

mzanetti:
Is this still waiting for feedback for real or just hasn't been updated yet?

mirko:
I'd be also *very* happy about this - e.g. librespot starts with 50% by default.
Just a config option which defines the default startup volume level would be fine (for now) - in case restoring volume is more / too much effort.

Mxbonn:
@SirWindfield any updates on this?

SirWindfield:
Currently in finals so I don't have time to work on it, sorry.

mzanetti:
Either spotifyd will kill my cats or they will kill me eventually, as every morning music starts playing at 100% volume and the whole house is shaking until I managed to find the slider in the app 😂😂😂

wilbertduenk:
You could set the normalisation_pregain to -10 or -20 or something to lower the maximum volume that spotifyd will play. Little workaround but it does the trick for me.

sflip:
I implemented `initial_volume` as a config option [here](#623).

SirWindfield:
Merged #623.

PR #623, opened 2020-06-25
feat(config): add initial volume

sflip:
This adds a config option `initial_volume` to set the initial volume, so it's not set to 100% [which may cause little kittens or their owners to die](https://github.com/Spotifyd/spotifyd/issues/445#issuecomment-604911070).

The setting is of course optional to be backwards compatible, so if it's not set, volume initialization isn't changed. Also, if the value could not be parsed, it prints a warning to help users and gently ignores it.

(This is my first contribution, please be gentle with me :slightly_smiling_face:)


mirko:
Can't tell you how glad I am about this feature. Working like a charme, finally switching from librespot. Thanks!

mzanetti:
Just managed to update my machine. Thank you so much!

I'm wondering a bit tho, wouldn't it make (more) sense to store the last used volume and restore that? That way a user wouldn't need to touch the config file at all. Seems to be what pretty much all other speakers/media players are doing.

SirWindfield:
It should actually do that already. Last time I tinkered around with the library I set it to cache the audio levels. There should be a json file inside the cache directory that contains the value. Not sure what broke it though.

sflip:
@mzanetti Agreed, restoring the last volume would be a lot better! This config feature can be considered a quickwin that seemed easier to implement in order to "fix" #445. 

@SirWindfield That's interesting! Really unfortunate that it broke. Maybe I'd like to look into that as well. But would that be the source of `librespot` rather than `spotifyd`, then?

SirWindfield:
I think so, yes.

PR #638, opened 2020-07-14
fix: Cannot start playback on device running spotifyd without a Spoti…

### LINKED ISSUE #609: Cannot start playback on device running spotifyd without a Spotify client
**Description**
Cannot start playback on device running spotifyd without another Spotify client.

It seems running spotifyd doesn't select the device as a playback device. Possibly the protocol requires a command to pick it as playback device.

**To Reproduce**
1. Start `spotifyd --no-daemon --username ... --password_cmd ...`
2. Run `sp wonderwall`
  2.1. Outcome: `Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.`
3. Open a Spotify client on local network
4. Run `sp wonderwall`
  4.1. Outcome: nothing
5. Show available devices on Spotify client from step 3
6. Run `sp wonderwall`
  6.1. Outcome: playback starts on Spotify client
7. Select the spotifyd device
8. Run `sp wonderwall`
  8.1. Outcome: playback starts on spotifyd client

**Expected behavior**
Be able to start playback on spotifyd through DBUS MPRIS (steps 1-2).

**Logs**
<details><summary>Click to show logs</summary>

<STEP 1>
jun 12 16:49:53: No config file specified. Running with default values
jun 12 16:49:53: No password specified. Checking password_cmd
jun 12 16:49:53: Running "pass" using "/bin/bash"
jun 12 16:49:53: No proxy specified
jun 12 16:49:53: Using software volume controller.
jun 12 16:49:54: Connecting to AP "guc3-accesspoint-b-f29x.ap.spotify.com:443"
jun 12 16:50:00: Authenticated as "awerlang" !
jun 12 16:50:00: Country: "BR"
<STEP 2>
<STEP 3>
<STEP 4>
<STEP 5>
jun 12 16:51:52: Unhandled DBus message: (Signal, Some("/org/freedesktop/DBus"), Some("org.freedesktop.DBus"), Some("NameAcquired"))
jun 12 16:51:52: Unhandled DBus message: (Signal, Some("/org/freedesktop/DBus"), Some("org.freedesktop.DBus"), Some("NameAcquired"))
jun 12 16:51:53: Couldn't fetch metadata from spotify: Ok(None)
jun 12 16:51:54: Couldn't fetch metadata from spotify: Ok(None)
jun 12 16:51:54: Unhandled DBus message: (Error, None, None, None)
jun 12 16:51:55: Unhandled DBus message: (Error, None, None, None)
jun 12 16:52:08: Couldn't fetch metadata from spotify: Ok(None)
<STEP 6>
<STEP 7>
jun 12 16:56:10: Loading <Wonderwall - Remastered> with Spotify URI <spotify:track:5wj4E6IsrVtn8IBJQOd0Cl>
jun 12 16:56:13: <Wonderwall - Remastered> loaded
<STEP 8>
jun 12 16:56:26: Loading <Wonderwall - Remastered> with Spotify URI <spotify:track:5wj4E6IsrVtn8IBJQOd0Cl>
jun 12 16:56:28: <Wonderwall - Remastered> loaded

</details>

**Compilation flags** 
- [x] dbus_mpris
- [ ] dbus_keyring
- [ ] alsa_backend
- [ ] portaudio_backend
- [x] pulseaudio_backend
- [ ] rodio_backend

 - OS: openSUSE Tumbleweed
 - Spotifyd: v0.2.24
- cargo: 1.43.0

awerlang:
…fy client

update rspotify to 0.10.0
activate dbus as soon as authentication succeeds
correctly retrieve device by name
transfer playback to current device on playback

Fixes #609

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2244,7 +2244,7 @@ dependencies = [
 
 [[package]]
 name = "rspotify"
-version = "0.8.0"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "base64 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -2564,7 +2564,7 @@ dependencies = [
  "librespot 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "percent-encoding 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "rspotify 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rspotify 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.111 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_ini 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha-1 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


awerlang:
There's an issue with `self.dbus_future`: after the authentication expires (1 hour), the daemon fails with:

```
Caught panic with message: cannot poll Map twice
```

With my understanding of the code, since a new token would need to be obtained, a new dbus_server would need to be created... I'm not sure why that doesn't happen.
I don't know if that's on purpose (require a new DbusServer), but maybe it could handle refreshing the auth token gracefully?

SirWindfield:
@awerlang  The code forrefreshinf the token is already in there, it just seems (for whatever reason) that it doesn't work.

https://github.com/Spotifyd/spotifyd/blob/6568e64762846b92b13d757cf48c1292be11d8f6/src/dbus_mpris.rs#L80-L93

awerlang:
I fixed the panic, but then the next issue became replacing the dbus server. It didn't handled new requests while songs were playing. Once the song stopped, it would take new requests. The previous instance would need to be shutdown properly, which I didn't attempted, but possibly it would fix the issue.

Now I have a working implementation where I keep the same dbus server, but replace the api token. I may push these changes once I can verify it doesn't create other issues, but yeah, an overall on the dbus would be good.

awerlang:
I found two issues that need fixing, no matter what approach to replace the bearer token.

1) The OpenUri commands doesn't wait for the new token to arrive, so the first request after the first token expires is still unauthorized;
2) The macro doesn't seem to want the token replaced. I figured that by calling `create_dbus_server` this should happen seamlessly, it doesn't though.

Any ideas?

SirWindfield:
>     2\. The macro doesn't seem to want the token replaced. I figured that by calling `create_dbus_server` this should happen seamlessly, it doesn't though.

That's actually weird. I just checked the code and it should actually replace the token. Are you sure it is the same one?



>     1. The OpenUri commands doesn't wait for the new token to arrive, so the first request after the first token expires is still unauthorized;

Maybe the functions should check that the token is still valid. I am not familiar at all with that part of the code to be honest. I'd have to take a closer look myself.

awerlang:
> > ```
> > 2\. The macro doesn't seem to want the token replaced. I figured that by calling `create_dbus_server` this should happen seamlessly, it doesn't though.
> > ```
> 
> That's actually weird. I just checked the code and it should actually replace the token. Are you sure it is the same one?
> 

I tested again with Spotify Connect, and it does indeed uses a new access token. I was curious about expiration, it's the same 3600 seconds.

> > ```
> > 1. The OpenUri commands doesn't wait for the new token to arrive, so the first request after the first token expires is still unauthorized;
> > ```
> 
> Maybe the functions should check that the token is still valid. I am not familiar at all with that part of the code to be honest. I'd have to take a closer look myself.

It would be better if it could wait for the new token to arrive as well, or maybe prefetch a new token once expiration is close.

PS: Your reply went through so fast that I guess this page picked it up and didn't create the notification?! Or somehow I missed it.

Issue #677, opened 2020-09-16
Cargo install install error!


404-unf:
Error Log is

```
   Compiling crossbeam-queue v0.1.2
   Compiling rustc_version v0.2.3
error: failed to run custom build command for `alsa-sys v0.1.2`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installetbaBH/release/build/alsa-sys-02fb35bbe057fef6/build-script-
build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "`\"pkg-config\" \"--libs\" \"--cflags\" \"alsa\"` did not exit 
successfully: exit code: 1\n--- stderr\nPackage alsa was not found in the pkg-config search path.\nPerhaps you should add the
 directory containing `alsa.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'alsa\' found\n"', 
/home/terra9/.cargo/registry/src/github.com-1ecc6299db9ec823/alsa-sys-0.1.2/build.rs:4:38
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile `spotifyd v0.2.24`, intermediate artifacts can be found at `/tmp/cargo-installetbaBH`

Caused by:
  build failed

```

terra9:
Cloning and compiling and then also has same error.
Error Log is
```
   Compiling rand_hc v0.1.0
   Compiling rand_xorshift v0.1.1
   Compiling rand_isaac v0.1.1
error: failed to run custom build command for `alsa-sys v0.1.2`

Caused by:
  process didn't exit successfully: `/media/terra9/4C68B5E468B5CCC8/Projects/yasdb/spotifyd/target/release/build/alsa-sys-
02fb35bbe057fef6/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "`\"pkg-config\" \"--libs\" \"--cflags\" \"alsa\"` did not exit
 successfully: exit code: 1\n--- stderr\nPackage alsa was not found in the pkg-config search path.\nPerhaps you should add the
 directory containing `alsa.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'alsa\' found\n"',
[...]

PR #689, opened 2020-10-07
make readme clearer on installing from source

robinvd:
directly using cargo install is not a subheader of installing from source making it clearer the same steps for dependencies have to be taken when installing with cargo.

see https://github.com/Spotifyd/spotifyd/issues/677#issuecomment-704783042

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,15 +67,13 @@ To install the resulting binary, run
 cargo install --path . --locked
 ```
 
-### Installing with Cargo
-
-If you have `cargo` installed, you can directly install `spotifyd` by running:
+You can directly install `spotifyd` without cloning the repo by running:
 
 ```bash
 cargo install spotifyd --locked
 ```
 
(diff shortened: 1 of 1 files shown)


slondr (approved):

README.md
```diff
@@ -67,15 +67,13 @@ To install the resulting binary, run
 cargo install --path . --locked
 ```
 
-### Installing with Cargo
```
Plecra, line 70: I'd rather theep the header - it breaks the explanation up into sections because they're two separate methods.

```md
### Alternative: Compiling without sources
```

While not strictly true, I think this is the functional difference between `build` and `install`: The user doesn't need to prepare the source directory.

Plecra (changes requested):

README.md
robinvd: thats fair, i do think it should be a subsection of the part that talks about dependencies, so that would be 4 `#`

PR #596, opened 2020-06-02
Update librespot dependency

marcelbuesing:
Waiting for https://github.com/librespot-org/librespot/pull/485 to be merged and a new librespot version.

Furthermore I'm unsure about the naming of strings in the `PLAYER_EVENT`, are those coming from somewhere specifically, e.g. `change`, `start`, `stop` ? 

For the gapless part alternatively using `.. Default::default()`, would have the same result, but might might make future changes non breaking (good or bad).

__Edit by SirWindfield:__
Fixes #85.

PR #695, opened 2020-10-22
Update dependencies

JojiiOfficial:
Tried to update as much dependencies as possible without breaking. Some adjustments in the src files were made. Mpris works for me. Some versions can't be updated unless we change some bigger parts of the code.

Closes #688, #627, #624, #305, #596


slondr:
What's up with the stuff about `matches!` reported by https://github.com/Spotifyd/spotifyd/pull/695/checks?check_run_id=1294697342?

JojiiOfficial:
Oh yeah lint 'errors'. I can fix them.

robinvd (approved):

robinvd:
@JojiiOfficial the lint errors should be fixed in a seperate PR, if you could update the conflicts? i think just removing the 'fix lint errors' commit should fix it

JojiiOfficial:
Removing the commit wasn't enough since #746 replaced the 'match block' with a 'matches macro' so I fixed the conflicts manually.

robinvd (approved):

slondr:
Seems like this PR missed the removal of `failure' from Cargo.toml, which was merged yesterday

Issue #702, opened 2020-10-29
New release/future of releases


robinvd:
Now that we have a new config format, and multiple reports of users using the new format with the old binary. I think it would be a good idea to make a new release on crates.io

@SirWindfield what is currently the setup for releasing to crates.io and who is allowed to do that?

massdregs:
Where can we get the new config file?

slondr:
> Where can we get the new config file?

@massdregs It's in the readme

SirWindfield:
I usually made sure that CI passed and I tested it locally on my Ubuntu VM to make sure that I get audio with pulse audio. After that, I create a new release stating which changes have been made:
```
git tag -a v0.x.x
```

It opens the editor and I manually add the changes and commit save and quit the editor afterwards. Then just push the tag. The CI will take care of the rest. If everything works fine and the artefacts have been generated, follow up the above with 

```
cargo publish
```

This will publish the version to crates.io. 

So basically:

1) Make sure everything works.
2) Make sure to create a release commit like "Bump version to 0.x.x". Just take a look at the commit history.
3) Create a git tag against that commit.
4) Push the tag to the repo.
4.1) Wait for CI to finish (there shouldn't be any problems if master passed all tests though).
5) Publish to crates.io

Optional 6) It would be probably better to create a changelog file and follow https://keepachangelog.com/de/1.0.0/. A lot of people use spotifyd and having to go through the releases to know what changed is utterly inconvenient. In the case you guys decide to create a changelog file, make sure to push the changelog changes after the versions have been published and not in-between or before. 
Pushing the changelog changes afterwards and not in-between makes sure that the versions published do work and use the same git reference across GitHub and crates. 
I usually do not push the changelog beforehand as I use automated tooling for changelog generation that I wrote some months back (@conventional-changelog-rs). It's not complete right now, a cargo subcommand is missing.

<details><summary>More about the tooling I use</summary>
You'd have to follow conventional commits so the tool can extract all the information it needs. https://github.com/SirWindfield/git-cm helps with that. When publishing a new release, you'd then run `next-semver | cargo next` to set the next version. Create a bumping commit: `chore(release): bump version to 0.x.x` and push the commits plus a new tag. If everything published succesfully, I run `changelog` to generate the new changelog and push it as well using a separate commit: `chore: update changelog`. 

The tooling basically needs to be run after pushing the newest tag as it uses git tags to derive the newest version and create the changelog accordingly.

[...]

PR #714, opened 2020-11-11
chore: add changelog

robinvd:
as discussed in https://github.com/Spotifyd/spotifyd/issues/702#issuecomment-719088894

ping @SirWindfield

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,20 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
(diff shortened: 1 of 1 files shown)


JojiiOfficial (approved):

SirWindfield:
Yep, something like this looks good!

SirWindfield (approved):

SirWindfield:
Not sure why CI gets triggered though, I set the paths to only account for Rust related files:
https://github.com/Spotifyd/spotifyd/blob/133a1199f80c7c22f29a45226db3bb856b106a5b/.github/workflows/ci.yml#L3-L8

SirWindfield:
Ah, looks like it is a conflict with the protected branch rule and the fact that the filter is applied inside the workflow:
https://github.community/t/path-filtering-on-required-pull-request-checks/18402/3

Just ping me if you want to get this merged into master, as I can still merge due to admin rights.

robinvd:
@SirWindfield i think with you two approving, this can be merged.

PR #721, opened 2020-11-22
Proposal for changes systemd service

takov751:
I have made this pull request as a proposal to restrict spotifyd in a safe environment while running it on Linux and of course on embeded systems like RPI  with running it a system service as a restricted user with only access to audio and the network. I do hope you find it useful in long term.

### DIFF
--- a/contrib/spotifyd.service
+++ b/contrib/spotifyd.service
@@ -9,7 +9,7 @@ After=network-online.target
 [Service]
 User=spotifyd
 Group=spotifyd
-WorkingDirectory=/var/lib/spoty
+WorkingDirectory=/var/lib/spotifyd
 ExecStart=dbus-run-session /usr/sbin/spotifyd --config-path /var/lib/spotifyd/spotifyd.conf --no-daemon
 Restart=always
 RestartSec=12
--- a/contrib/spotifyd.service
+++ b/contrib/spotifyd.service
@@ -7,8 +7,8 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-User=spotify
-Group=spotify
+User=spotifyd
+Group=spotifyd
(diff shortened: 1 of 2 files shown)


contrib/spotifyd.service
```diff
@@ -7,7 +7,10 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-ExecStart=/usr/bin/spotifyd --no-daemon
+User=spotify
+Group=spotify
```
igorbernstein, line 11: These don't match the user created in README `spoty`

robinvd: maybe spotify or spotifyd would be a better name?

takov751: I have seen your comments just a moment ago . I will recreate this pull request and correct it straight away

slondr:
Changing the systemd service file would be a breaking change for any automated installations, so this deserves careful consideration. Also, I'd like to see some copy-editing on the prose, although I'd be willing to do that myself if y'all are willing to wait until holidays are over :smile: 

I like the spirit and intention of the change, though. I think this is a good idea

contrib/spotifyd.service
```diff
@@ -7,7 +7,10 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-ExecStart=/usr/bin/spotifyd --no-daemon
+User=spotifyd
+Group=spotifyd
+WorkingDirectory=/var/lib/spoty
```
l4l, line 12: forgot to fix that line as well?

```suggestion
WorkingDirectory=/var/lib/spotifyd
```

takov751: Indeed . Thank you

takov751 (commented):
Corrected thank you

Issue #730, opened 2020-12-08
Gracefully handling dbus init errors


robinvd:
**Description**
currently when no x-server is available (or other dbus errors) we just panic, but this could just be a warning and then continue without dbus.

**To Reproduce**
1. compile with dbus enables d
2. run without an x-server (i thing unsetting DISPLAY should also work).
3. run spotifyd normally
4. crashes with a panic

**Expected behavior**
a warning, but not a panic

**Logs**
see https://github.com/Spotifyd/spotifyd/issues/729 for a relevant issue.

**Compilation flags** 
- [x] dbus_mpris
- [ ] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [ ] pulseaudio_backend
- [ ] rodio_backend

a quicker solution might be to provide a command line option to disable dbus.

relevant code locations:
- creating a dbus instance https://github.com/Spotifyd/spotifyd/blob/8cd1240eaa795adbb4fd7f5a55640312cf002741/src%2Fmain_loop.rs#L182
  this could be put behind an if triggered by a command line option, thus avoiding the panic.
- https://github.com/Spotifyd/spotifyd/blob/a863582fd1db456558fb55c4ac10835275afa8a9/src%2Fdbus_mpris.rs#L175
  this is where the panic happens, instead there could be a logger warning, and return something else

PR #740, opened 2020-12-21
Introduce use_mpris configuration setting

valir:
#730 

This defines the `use_mpris` configuration setting in the main
configuration file. Set this to `false` in case you have a headless
`spotifyd` instance, running in session lacking a dbus session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,9 +200,10 @@ use_keyring = true
 
 #
 # If set to true, `spotifyd` tries to bind to the session dbus
-# and expose MPRIS controls
+# and expose MPRIS controls. When running headless, without a dbus session,
+# then set this to false to avoid binding errors
 #
-use_mpris = false
+use_mpris = true
 
 # The audio backend used to play the your music. To get
 # a list of possible backends, run `spotifyd --help`.
--- a/README.md
+++ b/README.md
@@ -198,6 +198,12 @@ password_cmd = "command_that_writes_password_to_stdout"
 # can't be used simultaneously.
 use_keyring = true
 
+#
+# If set to true, `spotifyd` tries to bind to the session dbus
+# and expose MPRIS controls
+#
+use_mpris = false
+
 # The audio backend used to play the your music. To get
 # a list of possible backends, run `spotifyd --help`.
 backend = "alsa"
--- a/src/config.rs
+++ b/src/config.rs
@@ -623,14 +623,11 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
         .unwrap_or(DeviceType::Speaker)
         .to_string();
 
-    let pid = config
-        .pid
-        .map(|f| {
-            f.into_os_string()
-                .into_string()
-                .expect("Failed to convert PID file path to valid Unicode")
-        })
-        .or_else(|| None);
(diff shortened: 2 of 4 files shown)


valir:
Looks like the linter error is not in the code I've touched. Running the same `clippy` command on the master branch gives the same error.

valir:
Well, I've tried to come-up with this new feature and now I'm dragged down by errors with the linter on the code I don't touched and that are not reproducible on my laptop. Looks like the master code-base is not clean from the linter PoV. Please help me finish-up this pull request!

robinvd:
Hey @valir thanks so much for the PR on the issue i made! ill quickly check out the code and test it.

as for the CI other PRs also have this issue and i dont think it should be adressed here, when we fix it you can rebase/merge dev into this branch and the lint errors should go away. Sorry for the inconvience

README.md
```diff
@@ -198,6 +198,12 @@ password_cmd = "command_that_writes_password_to_stdout"
 # can't be used simultaneously.
 use_keyring = true
 
+#
+# If set to true, `spotifyd` tries to bind to the session dbus
+# and expose MPRIS controls
+#
+use_mpris = false
```
robinvd, line 205: i think keeping the current behaviour might be best, so default to using dbus

robinvd (changes requested):
code looks good, and quickly testing it seems to work great!

README.md
robinvd: Sorry i didnt explain it correctly, as it currenly is The Default impl of the settings (without a config) will set the value to false. And that would be a breaking change to existing users. I think the best solution would be to make it a `Option<bool>` then when setting the interal config do `config.shared_config.use_mpris.unwrap_or(true)`

robinvd (changes requested):
the value should also be merged by the configs, at https://github.com/Spotifyd/spotifyd/pull/740/files#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR517

valir:
After fetching upstream and rebasing on that master, I'm still getting `lint` errors in code I did not touched.

robinvd:
Yea that is unfortunately currently an issue. But we can just merge it even if there are lint errors

README.md
robinvd: Looks great!

valir:
Looking forward for the merge as I'd like to see it landing on Archlinux Arm repos 😉

robinvd:
@valir maybe i missed it but did you adress this comment already? https://github.com/Spotifyd/spotifyd/pull/740#pullrequestreview-566350806

valir:
> @valir maybe i missed it but did you adress this comment already? [#740 (review)](https://github.com/Spotifyd/spotifyd/pull/740#pullrequestreview-566350806)

The link in there shows an overview of the diffs my branch brings-in. Looks like I don't get what should be done there.

robinvd:
oh yeah sorry about that it should have gone to a specific line, the function config.rs:SharedConfigValues:merge_with merges the config from disc and from the command line. If you could add the new field so that it is also merged

robinvd (approved):

JojiiOfficial (approved):

PR #742, opened 2020-12-22
ci: fix needs clauses in steps

SirWindfield:


SirWindfield:
In theory this should trigger all steps...

PR #750, opened 2021-01-07
Added alternative to disable changing volume

Icelk:
I added an alternative to "softvol" and "alsa"; "none".
If remotely changing volume, or resetting it on restart is undesired, this is a fix.

### DIFF
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -32,6 +32,11 @@ pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLo
                 Box::new(|| Box::new(mixer::softmixer::SoftMixer::open(None)) as Box<dyn Mixer>)
                     as Box<dyn FnMut() -> Box<dyn Mixer>>
             }
+            config::VolumeController::None => {
+                info!("Using no volume controller.");
+                Box::new(|| Box::new(crate::no_mixer::NoMixer::open(None)) as Box<dyn Mixer>)
+                    as Box<dyn FnMut() -> Box<dyn Mixer>>
+            }
             _ => {
                 info!("Using alsa volume controller.");
                 let linear = matches!(
@@ -53,10 +58,17 @@ pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLo
     };
 
     #[cfg(not(feature = "alsa_backend"))]
-    let mut mixer = {
-        info!("Using software volume controller.");
-        Box::new(|| Box::new(mixer::softmixer::SoftMixer::open(None)) as Box<dyn Mixer>)
-            as Box<dyn FnMut() -> Box<dyn Mixer>>
+    let mut mixer = match config.volume_controller {
+        config::VolumeController::None => {
+            info!("Using no volume controller.");
+            Box::new(|| Box::new(crate::no_mixer::NoMixer::open(None)) as Box<dyn Mixer>)
+                as Box<dyn FnMut() -> Box<dyn Mixer>>
+        }
+        _ => {
+            info!("Using software volume controller.");
+            Box::new(|| Box::new(mixer::softmixer::SoftMixer::open(None)) as Box<dyn Mixer>)
+                as Box<dyn FnMut() -> Box<dyn Mixer>>
+        }
     };
 
     let cache = config.cache;
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -21,23 +21,18 @@ use log::info;
 use std::str::FromStr;
 
 pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLoop {
-    #[cfg(feature = "alsa_backend")]
     let mut mixer = {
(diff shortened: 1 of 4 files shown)


slondr (dismissed):

Iselk:
Is this a good solution or should I investigate if we can disable remote changing of volume? Maybe change the mixer in librespot to an Option<Mixer>? This just ignores input.
Perhaps the best way to move forward is to implement this and later change the librespot API?

Iselk:
What do you @JojiiOfficial @robinvd @SirWindfield think about this?

Iselk:
Does anybody know why the lint check is not successful?

Icelk:
Still waiting for response.

Icelk:
Still waiting for response.

slondr:
Looks like all the checks are passing now.

Icelk:
I’m a bit confused now.
I recently saw in `spotify-tui` the mobile clients have volume disabled. They advertised it to the Spotify API, and `spotify-tui` recognises it.

Should we merge this and then use the same struct for disabling it properly (if it’s possible)? If that’s the case, I think librespot needs to support the feature.

eladyn:
> I recently saw in `spotify-tui` the mobile clients have volume disabled. They advertised it to the Spotify API, and `spotify-tui` recognises it.
> 
> Should we merge this and then use the same struct for disabling it properly (if it’s possible)? If that’s the case, I think librespot needs to support the feature.

@Icelk While looking over the code today, I found by chance the following lines:

https://github.com/Spotifyd/spotifyd/blob/fbcbeca6562f02f53ab5e9771f7d8c2a3c9ae3cf/src/setup.rs#L69-L80

I was wondering, why this `struct` exists, since the actual volume control is done by the mixer, which has nothing to do with this setting. Apparently, this exists to signal to Spotify clients the type of volume control this device implements. And, most importantly, it has three possible states: `Log`, `Linear` and `Fixed`. I think the latter could be quite interesting to us. 

If you're still interested after all this time, this PR has been lying around, I suspect that it might be possible to turn off the client volume control via this setting! If not, that's completely understandable and okay too, of course!

Icelk:
Thanks for the consideration! You were completely correct.

I rewrote this PR to be up to date with the latest master, and implemented your idea. Thanks for the tip!

Icelk:
This PR now doesn't seem to have any issues. Can I improve anything @eladyn ?

src/setup.rs
```diff
@@ -53,10 +58,17 @@ pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLo
     };
 
     #[cfg(not(feature = "alsa_backend"))]
-    let mut mixer = {
-        info!("Using software volume controller.");
-        Box::new(|| Box::new(mixer::softmixer::SoftMixer::open(None)) as Box<dyn Mixer>)
-            as Box<dyn FnMut() -> Box<dyn Mixer>>
+    let mut mixer = match config.volume_controller {
```
eladyn, line 61: Instead of duplicating code here once more (between the `alsa_backend` and non-`alsa_backend` block), I _think_ the following structure should work as well?

```rust
let mut mixer = match config.volume_controller {
    config::VolumeController::None => {
        todo!("no volume controller")
    }
    #[cfg(feature(alsa_backend))]
    config::VolumeController::Alsa | config::VolumeController::AlsaLinear => {
        todo!("alsa volume controller")
    }
    _ => {
        info!("Using software volume controller");
        todo!();
    }
};
```

src/setup.rs
```diff
@@ -66,18 +78,25 @@ pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLo
...
+
     #[cfg(feature = "alsa_backend")]
     let volume_ctrl = if matches!(
         config.volume_controller,
         config::VolumeController::AlsaLinear
     ) {
         VolumeCtrl::Linear
     } else {
-        VolumeCtrl::default()
+        default_volume_ctrl
```
eladyn, line 95: If my above suggestion works, we should be able to merge these sections similarly as well.

eladyn (commented):
Wow! I wouldn't have expected such a fast response after such a long time. Thank you for the effort.

Just some comments on the code that might remove some duplicate code. The rest looks fine!

src/setup.rs
Icelk: True!

src/setup.rs
Icelk: That looks much better! Just implemented it.

Icelk:
> Wow! I wouldn't have expected such a fast response after such a long time. Thank you for the effort.

No problem :)

> Just some comments on the code that might remove some duplicate code. The rest looks fine!

Just pushed some fixes.

Thanks for your swift responses!

eladyn (approved):
LGTM and is working great. Thank you!

Icelk:
Thanks!

slondr (approved):
very awesome, thanks for the work on this!

PR #759, opened 2021-01-12
Remove `failure' as a dependency

slondr:
Refer to https://github.com/Spotifyd/spotifyd/pull/627#issuecomment-709347897; `failure` was not actually used, and is deprecated anyway, so this MR removes references to it in Cargo.toml and Cargo.lock.

Closes #627.


robinvd:
Weird that it is included, but good catch!

robinvd (approved):

JojiiOfficial (approved):

Issue #765, opened 2021-01-13
Spotifyd crashes after update (to 0.3.0)


wazabees:
---
Hijacking for visibility: Make sure to update to the new TOML config file format (see README) when using 0.3.0!!!
<sub>Edited by @SirWindfield</sub>

---

**Description**
spotifyd does not start after updating from version 0.2.24 to 0.2.25. Downgrading to 0.2.24 works.

**Expected behavior**
For spotifyd to launch in the background withour errors.

**Logs**
<details><summary>Click to show logs</summary>

journalctl log

```
Jan 13 09:06:40 arch systemd[945]: Started A spotify playing daemon.
Jan 13 09:06:40 arch spotifyd[10656]: Loading config from "/home/username/.config/spotifyd/spotifyd.conf"
Jan 13 09:06:40 arch spotifyd[10656]: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: NumberInvalid, line: Some(2), col: 11, at: Some(49), message: "", key: [] } }', src/config.rs:468:67
Jan 13 09:06:40 arch spotifyd[10656]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jan 13 09:06:40 arch systemd[945]: spotifyd.service: Main process exited, code=exited, status=101/n/a

```

Disabling the systemd unit and running manually:

```
╰─❯ spotifyd --verbose
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: NumberInvalid, line: Some(2), col: 11, at: Some(49), message: "", key: [] } }', src/config.rs:468:67
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

  

</details>

**Versions (please complete the following information):**

- OS: Arch linux, 5.10.6-arch1-1
- Spotifyd: 0.2.25
- cargo: 1.49.0

Comments:
The Arch linux package has a bug report here, that looks like the same issue:
https://bugs.archlinux.org/task/69308?project=5&string=spotifyd

The bug report on Arch also points to #719

JojiiOfficial:
Thats probably an issue which will affect most of the upgraded clients. Upgrade your config file to the format shown on the readme. We have added a breaking change to this release so parsing configs will fail if you used the old version in combination with a config file.

wazabees:
[...]

Issue #719, opened 2020-11-20
Panic: zero-initalisation of ov_callbacks


ChildOfDreams:
**Description**

I had successfully compiled and used Spotifyd for several days prior to this issue. I then decided I wanted to enable additional feature flags, so uninstalled the original instance and built again. 

However, after installing this new instance, I receive the following log messages when trying to play music from a Connect-capable client:

```Loading config from "/home/USERNAME/.config/spotifyd/spotifyd.conf"

No proxy specified

Using software volume controller.

Connecting to AP "gew1-accesspoint-b-qzv4.ap.spotify.com:443"

Authenticated as "USERNAME" !

Country: "GB"

Unhandled DBus message: (Signal, Some("/org/freedesktop/DBus"), Some("org.freedesktop.DBus"), Some("NameAcquired"))

Unhandled DBus message: (Signal, Some("/org/freedesktop/DBus"), Some("org.freedesktop.DBus"), Some("NameAcquired"))

Loading <Broken> with Spotify URI <spotify:track:6qF9QltwbDeujrVrzpCoLj>

Caught panic with message: attempted to zero-initialize type `librespot_tremor::tremor_sys::ov_callbacks`, which is invalid

Caught panic with message: called `Result::unwrap()` on an `Err` value: "SendError(..)"

Player thread panicked!
```
The application then exits.

Upon realising this, I attempted to rebuild the source with the original options and reinstall that, however the same error displays as above. I tried a variety of build options and appropriate configuration tweaking to no end, including using the crates.io release to no avail.

**To Reproduce**
Not something I can describe to reproduce, as it occurred suddenly and is persistent among rebuilds.

**Logs**
<details><summary>Click to show logs</summary>

  
Loading config from "/home/pi/.config/spotifyd/spotifyd.conf"

CliConfig { config_path: None, no_daemon: true, verbose: true, pid: None, shared_config: SharedConfigValues { username: Some("taken out for privacy"), password: Some("taken out for privacy"), password_cmd: None, use_keyring: false, on_song_change_hook: None, cache_path: None, no-audio-cache: false, backend: Some(PulseAudio), volume_controller: None, device: None, control: None, mixer: None, device_name: Some("RaspberryPiTV"), bitrate: Some(Bitrate320), volume_normalisation: false, normalisation_pregain: None, zeroconf_port: None, proxy: None } }

Found shell "/bin/bash" using SHELL environment variable.

No proxy specified

registering with poller

registering with poller

registering with poller

[...]

PR #769, opened 2021-01-14
also pin rust version on CI

robinvd:
in reply to this comment https://github.com/Spotifyd/spotifyd/issues/719#issuecomment-760148418

tested on my [fork](https://github.com/Spotifyd/spotifyd/runs/1701099531) and says it now uses the correct version, just like the last one this PR should also be reverted when it is fixed.

After this we can make a PR to push a new version/tag with 0.3 (https://github.com/Spotifyd/spotifyd/issues/765#issuecomment-760146839)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         build_target: [macos, linux, linux-armhf, linux-armv6]
-        rust: [stable]
+        rust: [1.47.0]
         artifact_type: ['slim', 'default', 'full']  # The build strategy will build all types for each OS specified
         include:
           - artifact_type: 'slim'               # Slim version has no features enabled by default.


JojiiOfficial (approved):

SirWindfield:
I actually thought about this and was sure that it would pick up the toolchain file but it seems it doesn't.

SirWindfield (approved):

robinvd:
I thought that as well, but after looking at the cd file it makes sense, as the toolchain is installed before the repo is downloaded

SirWindfield:
Make sure to also yank 0.2.25 and update the changelog with notable changes for 0.3 (config for example)! :)

robinvd:
i have yanked the crates.io release and make a new PR for 0.3.0

patricksamson:
None of the MacOS builds for v0.3.0 seem to be working for me. This is the error I'm getting : 
```bash
zsh: exec format error: ./spotifyd
```
It's as if the build wasn't executable. I also tested with 0.23.0 and 0.24.0 but those work fine so I don't think it's my machine.

SirWindfield:
It's the first time I've seen this kind of errors in general. Tried it on my laptop as well:
```$ ./spotifyd
Failed to execute process './spotifyd'. Reason:
exec: Exec format error
The file './spotifyd' is marked as an executable but could not be run by the operating system.
```

robinvd:
Maybe good for a separate issue?

in any case, thats quite weird and we havent really changed ci between 0.2 and 0.3

PR #778, opened 2021-01-16
fix inital_volume being a string

JojiiOfficial:


robinvd:
unfortunately this is a breaking change, we could also change the readme to `initial_volume = "90"` and keep the weird beheviour but that isnt great either

JojiiOfficial:
That is true but we don't need to publish a release instantly. Most people will use the latest stable version and this new way of parsing the "initial_value" will probably come with other minor breaking changes in the next version.

robinvd:
yes that might be a good plan, the only thing just like the TOML pr is that people will copy the config from the readme, and that wont be compatible

slondr:
I added this to the next breaking change milestone.

We should probably decide if we want to keep the incorrect documentation until then.

JojiiOfficial:
We should update the documentation to work with the currently stable version and update it when we release the new version. We could probably just add a comment that with a git version of spotifyd, the value must be set differently.

robinvd:
Or we could point the default branch to the latest release

JojiiOfficial:
Yes I was thinking about this as well. Usually I'm not a fan of such a solution because it looks like nothing is developed anymore and most people see the master branch as nightly/latest version (as far as I noticed) but in such a case like this, it is probably a solution we should consider.

slondr:
Or we could just hold off on merging this until we're ready to release 0.4.

SirWindfield:
Ye, just keep the PR open and merge once `v0.4` is on the horizon. It's the easiest.

leonm1:
I noticed this and wanted to bring up [changing the default branch](https://docs.github.com/en/github/administering-a-repository/changing-the-default-branch) to point to v.0.3.0, whereby drive-by readme readers could be shown the readme for the stable release and anyone looking to compile the master branch could do so with a simple `git switch master`. A note could be added to the readme for anyone who doesn't immediately notice that the default branch points to a stable release.

That way, breaking changes could be included with a consistent readme in `master` for people who want to build the current development branch, leading to more bleeding-edge users successfully testing the development channel and finding documentation errors and bugs (especially those arising from merging multiple breaking changes all together at release time) before release.

slondr:
Alternatively, do we publish the book anywhere? It would probably make more sense if we had properly versioned documentation without expecting users to realize the cargo.toml is a year out-of-date, lol

eladyn:
I'm not sure if you mean that, but the book is published [here](https://spotifyd.github.io/spotifyd/Introduction.html). It seems, however, that it is updated together with the master branch. Maybe we could

- either maintain two versions of it: e.g. https://spotifyd.github.io/spotifyd/master and https://spotifyd.github.io/spotifyd/latest
- or change that to only update on releases

rnestler:
One could change this to be an enum with `String` and `u16` to smooth out the transition: Warn if the user provided a string but still allow it. This would allow users to smoothly change to `u16` for the value and in the next release support for `String` can get removed.

PR #789, opened 2021-01-21
add better error reporting

robinvd:
use color_eyre to display errors, instead of just panicing and hoping the user will read through the long error to find whats wrong
this initial commit only has better errors for config loading as that is propably the most common error

for example:
```
Loading config from "/home/robin.local/.config/spotifyd/spotifyd.conf"
Error: 
   0: could not load the config file
   1: invalid type: integer `90`, expected a string for key `global.initial_volume` at line 8 column 18

Location:
   src/config.rs:478

note:
   the config format should be valid TOML
   we recently changed the config format, check the README for more info

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```


src/main.rs
```diff
@@ -59,8 +60,18 @@ fn main() {
...
-    cli_config.load_config_file_values();
+    color_eyre::install().expect("Coundn't initialize error reporting");
+
+    cli_config
+        .load_config_file_values()
+        .wrap_err("could not load the config file")
+        .with_section(|| {
+            concat!(
+                "the config format should be valid TOML\n",
+                "we recently changed the config format, check the README for more info"
```
JojiiOfficial, line 71: Looks fine! The only change I'd come up with would be to add a link to the certain github issue(s) here. That would make it easier to troubleshoot the known config issue(s). What do you think?

robinvd: this is the way color-eyre provides a stack of context, from the actual error to what that error means in the application. I can see how it looks like 2 errors, ill look if the lib supports some customization otherwise maybe we should remove it

SirWindfield:
It's even `tracing`-compatible, which makes it even great! The error message does look pretty good.

src/main.rs
robinvd: mm i cant seem to change the `Error:` to something like `Caused by:`, but removing the line `could not load the config file` would make it less obvious that the error came from loading the config file. So i think this is the best for now

robinvd: > The only change I'd come up with would be to add a link to the certain github issue(s) here

there is even some support for that (https://docs.rs/color-eyre/0.5.10/color_eyre/config/struct.HookBuilder.html#method.issue_url).

robinvd:
after adding github url:
```
Loading config from "/home/robin.local/.config/spotifyd/spotifyd.conf"
Error: 
   0: could not load the config file
   1: invalid number at line 2 column 15

Location:
   src/config.rs:478

note:
   the config format should be valid TOML
   we recently changed the config format, see https://github.com/Spotifyd/spotifyd/issues/765

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

JojiiOfficial (approved):

SirWindfield (approved):

robinvd:
A quick patch release might be a good idea, as most users might benefit from this improved error

JojiiOfficial:
Sounds good!

PR #790, opened 2021-01-21
add backticks to the bug report

robinvd:
this is much more readable in bug reports


SirWindfield:
Ye, I was hoping more people would use this but from my experience, I think only a handful did actually :(

SirWindfield (approved):

robinvd:
i did just see a typo in the original (and new) file `betwenn`

JojiiOfficial:
Oh yeah right didn't notice

PR #894, opened 2021-02-08
Update systemd documentation and systemd examples

tomodachi:
Hi people

It's not clear from the systemd  example exactly how a user systemd service works. 

With the given example spotifyd will only be started when the user logs in to the system
and  it will also be stopped as soon as the user logs out.

That might not be a expected behaviour for a setup  like a small headless device always connected to your speakers. 
So I've added an example for that and made some clarifications.


docs/src/config/services/Systemd.md
```diff
@@ -0,0 +1,35 @@
+# Running spotifyd as a systemd service
+
+## As as a user service
+
+Packagers of systemd-based distributions are encouraged to include the file in the former location. End-user should prefer the latter. It should be noted that some targets are not available when running under the user directory, such as `network-online.target`.
```
robinvd, line 5: This line is duplicated

robinvd (changes requested):
Thanks for the PR! more/better docs are always welcome.

docs/src/config/services/Systemd.md
```diff
@@ -0,0 +1,35 @@
+# Running spotifyd as a systemd service
```
eladyn, line 1: The typo in the filename should also be corrected in the [SUMMARY.md](https://github.com/Spotifyd/spotifyd/blob/603bcc4d802d2b8f45ac517c0513eb4096252d7b/docs/src/SUMMARY.md?plain=1#L15) file,

docs/src/config/services/Systemd.md
```diff
@@ -0,0 +1,35 @@
...
+A `systemd.service` unit file is provided to help run spotifyd as a service on systemd-based systems. The file `contrib/spotifyd.service` should be copied to:
+
+```bash
+/etc/systemd/system/
+```
+
+Control of the daemon is handed over to systemd. The following example commands will start the service and keep it running across reboots.
+
+```bash
+systemctl -daemon-reload
```
eladyn, line 33:
```suggestion
systemctl daemon-reload
```
Typo in the command. Also, I'm not sure, wether this is actually needed, since it isn't mentioned above either.

docs/src/config/services/Systemd.md
```diff
@@ -0,0 +1,35 @@
...
+systemctl --user enable spotifyd.service --now
```
eladyn, line 34:
```suggestion
systemctl enable spotifyd.service --now
```
Since this should enable the system-wide service, the `--user` flag isn't needed.

eladyn (changes requested):
I don't know, if you are still interested in this @tomodachi, but if you are (which would be great), I have some small suggestions. Other than that, LGTM.

docs/src/config/services/Systemd.md
eladyn: Already done in https://github.com/Spotifyd/spotifyd/commit/b46a3002c3f35b017d33d20104d02a92a2d6f6e5.

eladyn (approved):
LGTM!

slondr (approved):
cool

slondr:
@robinvd  this look good to you?

robinvd (approved):

PR #602, opened 2020-06-11
Windows support

JCapucho:
This PR allows `spotifyd` to run on windows as non-daemon and daemon-like, only the rodio-backend is supported and the config file is stored on:

```
%AppData%/Roaming/spotifyd/spotifyd.conf
```

All output is logged into a `.spotifyd.log` file on windows when running in both non-daemon and daemon mode although in daemon mode the startup is logged to the console before a detached subprocess with the same binary and arguments (with the exception that the argument `--no-daemon` is set) is started

PR #922, opened 2021-03-18
add windows support - see (#602)

frankmoskal:
See (#602) for details. All credit goes to  @JCapucho.

My changes are:
- rebasing against master
- making rodio the default background (platform-independent) & update windows-specific dependencies
- making the default config location based on [directories](https://docs.rs/directories/1.0.1/directories/struct.BaseDirs.html) without a fallback  + a bit of logging

Feel free to change whatever! I've really only tested compiling this with msvc, but it works for me.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -293,6 +293,11 @@ pub struct SharedConfigValues {
     )]
     password_cmd: Option<String>,
 
+    /// Whether the credentials should be debugged.
+    #[structopt(long)]
+    #[serde(skip)]
+    debug_credentials: bool,
+
     /// A script that gets evaluated in the user's shell when the song changes
     #[structopt(visible_alias = "onevent", long, value_name = "string")]
     #[serde(alias = "onevent")]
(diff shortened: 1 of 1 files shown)


frankmoskal:
I wanted to clean this up a little - will reopen at a later date.

Alexhuszagh:
@frankmoskal This doesn't seem to work: a great example is daemonize doesn't support Windows. There's a lot of imports that simply aren't supported on Windows, like `use std::os::unix`, and I'm not sure how to clean that up without a substantial rewrite.

Alexhuszagh:
Ok the dependencies that don't seem to work are as follows:

- syslog (a Unix-only system logger)
- dbus, dbus-tokio (optional, no big issue)
- alsa (optional, not used for our backend)
- daemonize

The main issue is daemonize, which seems to be unmaintained, Unix-specific, and may take some work to port over. syslog v5 supports Windows, while v4 and v4.0.1 only support Unix.

Alexhuszagh:
The closest example I can see to daemonize is daemon-rs, which should be a trivial replacement (if it works on non-Linux Unix-Like systems and macOS), since the only code requiring daemon support is:

```rust
    if is_daemon {
        info!("Daemonizing running instance");

        let mut daemonize = Daemonize::new();
        if let Some(pid) = internal_config.pid.as_ref() {
            daemonize = daemonize.pid_file(pid);
        }
        match daemonize.start() {
            Ok(_) => info!("Detached from shell, now running in background."),
            Err(e) => error!("Something went wrong while daemonizing: {}", e),
        };
    }
```

Alexhuszagh:
I'm unsure of how we'd port daemonize, since the original relies on fcntl for the Python source, which is Unix-specific, just like the port.

frankmoskal:
@Alexhuszagh You'll actually want to look at [frankmoskal-windows-dev](https://github.com/frankmoskal/spotifyd/tree/frankmoskal-windows-dev ) - I moved things over to that branch after closing this MR to make things a bit easier on my end.

Alexhuszagh:
@frankmoskal Ah my bad, I didn't see anything obvious in the branch list and just assumed it was on master.

Issue #654, opened 2020-08-05
Improve logging 


SirWindfield:
- [ ] make syslog optional
- [ ] look into tracing for easier async logging

ericonr:
I've seen a lot of users facing issues related to the lack of `syslog` in their systems.

I think at least improving the error message is the more urgent change, since otherwise it's a bit hard to diagnose it. (I ended up stracing it to see what syscalls were failing, and found the attempts to open `/dev/log` and `/var/run/syslog` - which was made more complicated by it only using syslog when backgrounding itself, so at first I couldn't reproduce it at all). Would a PR for this be accepted? It's not an ideal solution, but it is a quality of life improvement.

It might be nice to pin one of the issues below, so people who come to the repo have an easier time finding it. There are quite a few duplicates already.

Some issues: #644 #539 #576 #720 ; I also had people reach out on `#voidlinux` on Freenode.

Thanks!

ericonr:
This isn't stale.

PR #933, opened 2021-04-08
Created cli-option log-path that when set overrides syslog

fennewald:
Progress on #654, added a cli option `--log-path` that override syslog as a target for logging.

Also made the no-syslog error message a little clearer.


ericonr:
I guess a user who wishes to discard logs could send them to `/dev/null`?

ericonr:
This isn't stale.

Issue #948, opened 2021-05-08
[Question] More than one user


aanno:
We've got a (paid) spotify family account and I have set up spotifyd with my user. However with this, my son could not 'respot' to spotifyd (but he could see the device in his official spotify android app). Is this the expected behaviour? Is there a way around?

BTW, there a 'dump' devices like speakers that spotify could 'respot' to that are not registered/associated with a spotify account. I wonder if there is way to run spotifyd in a similiar mode?

eladyn:
This should AFAIK be possible by just omitting the `username`, `password`, `username_cmd` and `password_cmd` options. It at least works for me on 0.3.2.

aanno:
Yes, I tried to leave out the spotify account in the configuration, as the people at `librespot` said this is possible and `librespot` would work in 'discovery mode' then.

However:
```bash
$ systemctl --user status spotifyd
● spotifyd.service - A spotify playing daemon
     Loaded: loaded (/etc/xdg/systemd/user/spotifyd.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2021-05-17 08:51:45 CEST; 49s ago
       Docs: https://github.com/Spotifyd/spotifyd
   Main PID: 32070 (spotifyd)
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/spotifyd.service
             └─32070 /usr/bin/spotifyd --no-daemon

Mai 17 08:51:45 unke spotifyd[32070]: Loading config from "/etc/spotifyd.conf"
Mai 17 08:51:45 unke spotifyd[32070]: No username specified. Checking username_cmd
Mai 17 08:51:45 unke spotifyd[32070]: No username_cmd specified
Mai 17 08:51:45 unke spotifyd[32070]: No password specified. Checking password_cmd
Mai 17 08:51:45 unke spotifyd[32070]: No password_cmd specified
Mai 17 08:51:45 unke spotifyd[32070]: No proxy specified
Mai 17 08:51:45 unke spotifyd[32070]: Using software volume controller.
Mai 17 08:51:45 unke spotifyd[32070]: Connecting to AP "gew1-accesspoint-e-skbp.ap.spotify.com:443"
Mai 17 08:51:45 unke spotifyd[32070]: Authenticated as "<my_spotify_account_name>" !
Mai 17 08:51:45 unke spotifyd[32070]: Country: "DE"
```

[...]

Issue #938, opened 2021-04-18
can't see a machine running spotifyd on my "Duo account"


alembiq:
**Is your feature request related to a problem? Please describe.**
Im running spotifyd on my router which is used to play music in my kitchen & I have a Spotify Duo account.
Before when using RPI and [raspotify](https://github.com/dtcooper/raspotify), both my and my GF (the other account on my Duo subscription) were seeing the RPI as a speaker in the "devices to connect and listen on".

Both raspotify and spotifyd are configured to use my account, but both accounts are linked by the Duo subsription.

I would love to have the same feature here, for both of us being able to use spotifyd as a shared spotify device for both accounts in Spotify Duo Subscription.

**Describe the solution you'd like**
Same solution that is in [raspotify](https://github.com/dtcooper/raspotify).

eladyn:
In order to advertise the player and to be able to select it from any account, you can just leave `username`, `password`, `username_cmd` and `password_cmd` empty. Spotifyd will then accept anyone who tries to connect to it and will be visible to anyone on the same network.

Hope this helps!

alembiq:
nice, will try when i get back next week :) thanks

alembiq:
yeah, i forgot.....

nevertheless, even when i commented out `username` and `password` my fiancee cannot see the device in the network - checked both on mobile and computer :(
while i can see it both on my computer and mobile and both of us are sitting on the same network

it's making me confused

honza:
I'm having the same issue.

honza:
I solved the problem in my setup. I removed the username and password from `spotifyd.conf`, and then set the `zeroconf_port` to `1234`, and opened the port on the Linux machine that runs spotifyd.

PR #955, opened 2021-05-16
document zeroconf based authentication properly

eladyn:
In order to be able to use zeroconf based authentication, it is currently necessary to omit everything related to username or password in the configuration. 

AFAICT, this is not really documented anywhere except for [this wiki entry](https://spotifyd.github.io/spotifyd/config/index.html), which does barely go into details about this. Maybe I just haven't looked in the right places for this, but several others (#938, #948) seem to have trouble finding anything as well.

### DIFF
--- a/docs/src/config/File.md
+++ b/docs/src/config/File.md
@@ -105,7 +105,7 @@ device_type = "speaker"
 
   Spotifyd is able to advertise itself on the network without credentials. To enable this, you must omit / comment any `username` / `username_cmd` or `password` / `password_cmd` in the configuration. Spotifyd will receive an authentication blob from Spotify when you choose it from the devices list.
 
-  > __Note:__ If you choose to go with this, it is also recommended to omit the `cache_path` option. Otherwise the first user to connect to the service will have its authentication blob cached by the service and nobody else will be able to connect to the service without clearing the cache.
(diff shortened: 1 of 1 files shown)


eladyn:
I rebased this PR on the latest master. It would be great to get some feedback on this! :smiley:

tomodachi:
Yes plz merge.

Just enabled this and It works great  👍🏽 

Nice find @eladyn 
Much better than running my hackish solution of running multiple spotifyd daemons with different users on the same device.

eladyn:
Still relevant (I think).

mas-4:
@eladyn I commented in the relevant issue (https://github.com/Spotifyd/spotifyd/issues/948) and will repeat here: it seems you also need to have caching disabled for this to work. When cache_directory is set in the configuration the credentials of the first user to connect are cached and then anyone else who tries to connect causes the program to restart (using a systemd service with restart=always restartsec=12).

eladyn:
@mas-4 Thanks for investigating this! I added a small note about the caching behaviour. It would be great if you could check whether this fits with your findings.

As for the `no_credential_caching` thing, I will rather comment on the issue.

mas-4:
@eladyn I actually had `cache_directory` set which was also the problem. Might be worth noting both parameters.

eladyn:
I currently can't find any mention of `cache_directory`, neither in the docs nor in the actual [code](https://github.com/Spotifyd/spotifyd/blob/f4909621da43636185f2d2ffe4392178b8d654b6/src/config.rs#L306-L308). But it surely doesn't hurt to add it.

SimonPersson:
Thanks!

Issue #969, opened 2021-06-17
Tracking issue for upgrading librespot to v0.2.0


JojiiOfficial:
As mentioned in #717 here the tracking issue for upgrading librespot to v0.2.0

closes #717 #945

capnfabs:
Hey! I started work on this a few weeks ago for a hobby project ([video](https://capnfabs.net/posts/why-is-this-floppy-disk-drive-clicking-incessantly/ill-communication.mp4)); the upgrade work is by no means complete yet, but wanted to link it here in case it's helpful to someone as a reference for a cleaner PR. 

Current status: it works, but:
- it panic whenever you try to quit with Ctrl+C (this is probably easy to fix, I just haven't done it yet)
- dbus is a bit complicated. The dbus libraries changed significantly as part of supporting tokio 1.0, which is required for librespot v0.2.0. I'd wanted to rework the current dbus_mpris implementation anyway -- would be nice to support property change signals, and to not hit the Spotify API on every dbus property fetch -- so I've been doing it in the same branch so that I can use `dbus_crossroads`, which seems to be the "newer" way of implementing a DBUS server.
- I've ignored automated tests so far.


Code is here: https://github.com/Spotifyd/spotifyd/compare/master...capnfabs:fabian/wip-big-upgrade

Would you be open to reworking DBUS and doing the upgrade in the same PR? I would _normally_ try and do the minimal possible upgrade in a PR, but am very much _not_ doing this in the code on that branch.

capnfabs:
Cleaned up that branch a bit, fixed the Ctrl+C issue, and pushed it to https://github.com/Spotifyd/spotifyd/pull/977

PR #951, opened 2021-05-10
Use librespot libraries instead of its binary crate

hhirtz:
librespot's main crate is just a reexport of its sub-modules: https://github.com/librespot-org/librespot/blob/v0.1.5/src/lib.rs#L4

The number of dependencies on spotifyd can be reduced by depending on librespot's sub-modules instead, this PR does that.

I didn't bump librespot's version to the [latest 0.2][0] since it depends on tokio 1 instead of tokio 0.1.

Note: this PR has the side effect of doing a `cargo update` on the librespot dep

[0]: https://github.com/librespot-org/librespot/releases/tag/v0.2.0

PR #977, opened 2021-07-11
Upgrade librespot to 0.2.0

capnfabs:
This is an initial attempt at upgrading librespot to 0.2.0. That also required:
- upgrading tokio to 1.0
- reworking `dbus_mpris.rs` to use `dbus-crossroads`. 
 
The dbus-mpris stuff was all written using an older version of `dbus-rs`, which integrated with tokio 0.1, and therefore needed to be upgraded. I initially tried to do this with the `dbus-tree` package to minimise code changes, but I found it difficult to reason about the threading model, and eventually switched to dbus-crossroads. I hope that's ok!

I've left a number of TODOs in the code:
- some are product-ish questions (e.g. "do we want to support this new librespot feature in the spotifyd config?")
- some are places where the code works on my machine, but I'm unclear on whether it will work for everyone.

I'd love some input on how to resolve those!

Fixes https://github.com/Spotifyd/spotifyd/issues/969

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,24 +14,26 @@ daemonize = "0.4"
 dbus = { version = "0.6", optional = true }
 dbus-tokio = { version = "0.2", optional = true }
 fern = { version = "0.6.0", features = ["syslog-4"] }
-futures = "0.1"
+futures = "0.3.15"
 gethostname = "0.2.0"
 hex = "0.4"
 keyring = { version = "0.10.1", optional = true }
 libc = "0.2.82"
 log = "0.4.6"
 percent-encoding = "2.1.0"
+reqwest = "0.11.3"
 rspotify = "0.8.0"
 serde = { version = "1.0.115", features = ["derive"] }
 sha-1 = "0.9"
 structopt = "0.3.17"
 syslog = "4"
-tokio-core = "0.1"
-tokio-io = "0.1"
-tokio-signal = "0.1"
+tokio = "1.6.1"
+tokio-compat = { version = "0.1.6", features = ["rt-current-thread"] }
+tokio-compat-02 = "0.2.0"
+tokio-stream = "0.1.7"
 url = "1.7"
 xdg = "2.2"
-librespot = { version = "0.1.5", default-features = false, features = ["with-tremor"] }
+librespot = { version = "0.2.0", default-features = false, features = ["with-tremor"] }
 toml = "0.5.8"
 color-eyre = "0.5"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,9 @@ version = "0.3.2"
 alsa = { version = "0.3", optional = true }
 chrono = "0.4"
 daemonize = "0.4"
-dbus = { version = "0.6", optional = true }
-dbus-tokio = { version = "0.2", optional = true }
+dbus = { version = "0.9", optional = true }
+dbus-tokio = { version = "0.7.3", optional = true }
+dbus-crossroads = { version = "0.4.0", optional = true }
 fern = { version = "0.6.0", features = ["syslog-4"] }
 futures = "0.3.15"
 gethostname = "0.2.0"
@@ -46,7 +47,7 @@ env_logger = "0.7"
 [features]
 alsa_backend = ["librespot/alsa-backend", "alsa"]
 dbus_keyring = ["keyring"]
-dbus_mpris = ["dbus", "dbus-tokio"]
+dbus_mpris = ["dbus", "dbus-tokio", "dbus-crossroads"]
 default = ["alsa_backend"]
 portaudio_backend = ["librespot/portaudio-backend"]
 pulseaudio_backend = ["librespot/pulseaudio-backend"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "GPL-3.0-only"
 version = "0.3.2"
 
 [dependencies]
-alsa = { version = "0.3", optional = true }
+alsa = { version = "0.5", optional = true }
 chrono = "0.4"
 daemonize = "0.4"
 dbus = { version = "0.9", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,13 +28,16 @@ serde = { version = "1.0.115", features = ["derive"] }
 sha-1 = "0.9"
 structopt = "0.3.17"
 syslog = "4"
-tokio = "1.6.1"
+tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread"] }
 tokio-compat = { version = "0.1.6", features = ["rt-current-thread"] }
 tokio-compat-02 = "0.2.0"
 tokio-stream = "0.1.7"
 url = "1.7"
 xdg = "2.2"
-librespot = { version = "0.2.0", default-features = false, features = ["with-tremor"] }
+librespot-audio = { version = "0.2.0", default-features=false, features = ["with-tremor"] }
+librespot-playback = { version = "0.2.0", default-features=false }
+librespot-core = { version = "0.2.0"}
+librespot-connect = { version = "0.2.0"}
 toml = "0.5.8"
 color-eyre = "0.5"
 
@@ -45,13 +48,13 @@ whoami = "0.9.0"
 env_logger = "0.7"
 
 [features]
-alsa_backend = ["librespot/alsa-backend", "alsa"]
+alsa_backend = ["librespot-playback/alsa-backend", "alsa"]
 dbus_keyring = ["keyring"]
 dbus_mpris = ["dbus", "dbus-tokio", "dbus-crossroads"]
 default = ["alsa_backend"]
-portaudio_backend = ["librespot/portaudio-backend"]
-pulseaudio_backend = ["librespot/pulseaudio-backend"]
-rodio_backend = ["librespot/rodio-backend"]
+portaudio_backend = ["librespot-playback/portaudio-backend"]
+pulseaudio_backend = ["librespot-playback/pulseaudio-backend"]
+rodio_backend = ["librespot-playback/rodio-backend"]
 
 [package.metadata.deb]
 depends = "$auto, systemd, pulseaudio"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,36 +69,14 @@ dependencies = [
 
 [[package]]
 name = "alsa"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe6a5e4eb4879d7cb8d19161927fbf825f0d89761776359c9cf4f9c9c2f35c52"
-dependencies = [
- "alsa-sys 0.1.2",
- "bitflags 0.9.1",
- "libc",
- "nix 0.14.1",
-]
-
-[[package]]
-name = "alsa"
 version = "0.5.0"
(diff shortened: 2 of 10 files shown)


busimus:
I'm desperate for an upgrade to MPRIS so I tested this PR. Here's what I found:

1. spotifyd doesn't show up in `playerctl -a status` as it did previously and running `playerctl -p spotifyd play` returns "No player could handle this command", even though spotifyd shows up in `playerctl -l`. 
I tried changing the object path in code from `/` back to `/org/mpris/MediaPlayer2` like it was, didn't fix everything – `playerctl` commands started working but you still have to specify `-p spotifyd` or it will just ignore spotifyd and send the command to some other player. This used to work right.

2. It still takes an unreasonably long time to query basic properties like `PlaybackStatus`, which for me takes up to half a second to respond (when using `dbus-send`). This used to slow down `playerctl` a lot and it still does. With the path fix mentioned above, running `playerctl -p spotifyd play-pause` takes up to an entire second.

Oh and also enabling `volume_normalisation` completely blows out the sound, but I'm not sure whose fault is it.

Would love to help getting any of this debugged if needed, because otherwise the MPRIS feature will remain unusable if you wish to control multiple players, not just spotifyd.

capnfabs:
Hi Ryan!

Thanks for testing out this PR; this is all super helpful feedback. I'm
away from my computer until Wednesday and will take a look sometime
after I'm back.

For what it's worth - I totally agree with you about the DBUS situation
in general. The implementation here calls out to the Spotify web api for
almost every DBUS message (PlayerStatus and Metadata being two of the
most notable). I've tried to keep that the same in this PR in order to
limit the scope of the PR, but I've also been working on a fix for that.

The fix hooks into librespot's player events and metadata retrieval
functionality. It's made a huge improvement in my tests (no need for a
network call on DBUS message!) but requires some changes to the upstream
librespot dependency to work as currently implemented. I'll share the
code for that with you when I'm back too; if you're interested in
testing.

On July 24, 2021, Ryan Summers ***@***.***> wrote:
> I'm desperate for an upgrade to MPRIS so I tested this PR. Here's what
> I found:
>
> 1. spotifyd doesn't show up in playerctl -a status as it did
>    previously and running playerctl -p spotifyd play returns "No
>    player could handle this command", even though spotifyd shows up in
>    playerctl -l.
>     I tried changing the object path in code from / back to
>    /org/mpris/MediaPlayer2 like it was, didn't fix everything –
>    playerctl commands started working but you still have to specify -p
>    spotifyd or it will just ignore spotifyd and send the command to
>    some other player. This used to work right.
>
> 2. It still takes an unreasonably long time to query basic properties
>    like PlaybackStatus, which for me takes up to half a second to
>    respond (when using dbus-send). This used to slow down playerctl a
>    lot and it still does. With the path fix mentioned above, running
>    playerctl -p spotifyd play-pause takes up to an entire second.
>
>
> Oh and also enabling volume_normalisation completely blows out the
> sound, but I'm not sure whose fault is it.
>
> Would love to help getting any of this debugged if needed, because
> otherwise the MPRIS feature will remain unusable if you wish to
> control multiple players, not just spotifyd.
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Spotifyd/spotifyd/pull/977#issuecomment-885888881>,
or unsubscribe <https://github.com/notifications/unsubscribe-
auth/AAKGJPBDIANIQX4RVGISW53TZHEHRANCNFSM5AFRAILQ>.

capnfabs:
Hey! I tested this briefly today. Here are the results:

- Spotifyd doesn't show up in dbus until after a track starts playing. That's new as of this PR.
- The DBUS rework appears to entirely not work on Linux? I'm not sure why, I tested this extensively on OSX.

I'll dig into it. 

Haven't looked into the volume normalization yet; but I see log messages saying

> This track will at its peak be subject to NaN dB of dynamic limiting.

Which sounds like it might be the culprit; I might've rigged something up badly.

capnfabs:
Alright:

> The DBUS rework appears to entirely not work on Linux? I'm not sure why, I tested this extensively on OSX.

The problem here was indeed what @busimus mentioned here:

> changing the object path in code from / back to /org/mpris/MediaPlayer2 like it was


Re this:

> playerctl commands started working but you still have to specify -p spotifyd or it will just ignore spotifyd and send the command to some other player. This used to work right.

@busimus, what do you mean by "right" in this case? I'm not super familiar with dbus, but my understanding is that [playerctl chooses arbitrarily unless you have playerctld running](https://github.com/altdesktop/playerctl#selecting-players-to-control). I'm not sure how `playerctld` monitors player status, but I could imagine it interacting badly with the way DBUS is implemented at the moment (doesn't send change signals, gives weird results if the API call gets rate limited, which happens somewhat frequently).

Final DBUS regression:

> Spotifyd doesn't show up in dbus until after a track starts playing

This was something gnarly to do with futures / async/await that I don't quite understand, but I've added a hack to workaround it.

capnfabs:
Alright, fix for normalization is in too.

busimus:
> what do you mean by "right" in this case?

This was probably me forgetting that I wrote many workarounds to make it work right. I guess the dream is that `playerctl -p playerctld play-pause` command works fast and without additional hacks.
I tested everything again, the latest released version of spotifyd indeed does _not_ work right: playerctld doesn't notice changes in activity of spoytifyd so it ignores it after it discovers another player. 
With this PR, it ignores spotifyd completely, it doesn't show up in `playerctl -l`. Something I noticed when running playerctld in console is that it throws this warning when spotifyd starts:
```
(playerctld:322578): playerctl-WARNING **: 21:25:28.160: could not get properties for active player: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Path, Interface, or Method does not exist
```

> Alright, fix for normalization is in too.

I wish I could say I remembered to disable normalization when I started testing new changes before that commit. But it did fix it.

capnfabs:
FWIW, I think the issues with play-pause being slow to respond are actually not in the DBUS implementation; it's got something to do with the way frames are buffered in ALSA. I've been using the Rodio backend on OSX and it's been very responsive, at least.

There are additional issues with other MPRIS methods requiring a server round trip -- notably, the `metadata` command. I've reworked that in https://github.com/Spotifyd/spotifyd/compare/master...capnfabs:dbus-event-plumbing to use events from librespot, but if you want to test it out, you'll also need this fork of librespot: https://github.com/librespot-org/librespot/compare/dev...capnfabs:flopbox

busimus:
I'm using the pulseaudio backend and it stops fast enough if I use `dbus-send`. But `playerctl -p spotifyd play-pause` takes up to a second, orders of magnitude longer than any other player. 

I tested that branch. Getting metadata or playback status is instantaneous but strangely enough `playerctl -p spotify play-pause` still takes 0.25 seconds.
Looking through `dbus-monitor` output I found that playerctl called the `Properties.GetAll` method for the player and it caused the slowdown. Here's that method call as a command:
```
dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotifyd /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.GetAll string:org.mpris.MediaPlayer2.Player
```

So, if you manage to speed up `GetAll` and get `playerctld` to stop ignoring spotifyd with that warning, I think the dbus issue will be solved for good (given that all the changes get accepted into both projects of course).

capnfabs:
Ah, I think I know what's going on -- `GetAll` would also fetch the loop status and repeat properties; and those (and maybe Position?) are currently still retrieved through the Spotify API.

Thanks for the repro command, that's super helpful! Sounds like it might be worth opening an issue with librespot and asking them if they've got ideas for fetching Loop/Shuffle status, but I'm also inclined to say "let's not worry about it for this PR and do it in the future given that it's not a regression."

busimus:
> let's not worry about it for this PR and do it in the future given that it's not a regression

Yeah you're right. 
I changed shuffle and loop status properties to return constants (because I don't care about them) and now everything is fast. Still requires a little workaround script to make it work seamlessly, but it's good enough for me to use daily, thank you.

busimus:
I tested that debug/release mode race condition a bit. It happens in the exact same way both in this PR and in the other branch: in debug mode playerctld throws the warning and ignores spotifyd, in release mode warnings happen rarely and when they don't then playerctl works perfectly (but slowly in this PR). 
In all cases playerctld tries to `GetAll` properties for two interfaces and it fails if spotifyd takes too long to set those interfaces up after calling `conn.request_name(...)` (so adding a sleep after it makes it reproduceable).

In case you need them, here are `dbus-monitor` logs for all cases:
[spd_pr_debug.log](https://github.com/Spotifyd/spotifyd/files/6923671/spd_pr_debug.log)
[spd_pr_release.log](https://github.com/Spotifyd/spotifyd/files/6923672/spd_pr_release.log)
[spd_branch_debug.log](https://github.com/Spotifyd/spotifyd/files/6923673/spd_branch_debug.log)
[spd_branch_release.log](https://github.com/Spotifyd/spotifyd/files/6923674/spd_branch_release.log)

Cargo.toml
```diff
@@ -8,30 +8,33 @@ license = "GPL-3.0-only"
...
-tokio-io = "0.1"
-tokio-signal = "0.1"
+tokio = "1.6.1"
+tokio-compat = { version = "0.1.6", features = ["rt-current-thread"] }
+tokio-compat-02 = "0.2.0"
+tokio-stream = "0.1.7"
 url = "1.7"
 xdg = "2.2"
-librespot = { version = "0.1.5", default-features = false, features = ["with-tremor"] }
+librespot = { version = "0.2.0", default-features = false, features = ["with-tremor"] }
```
hhirtz, line 37: spotifyd should depend on librespot's individual crates (e.g. `librespot-audio` or `librespot-connect`), since the `librespot` crate is a binary and has a lot of dependencies spotifyd doesn't use, like `miow`, `getopts`, `rpassword` and `sockets2`.

See https://github.com/Spotifyd/spotifyd/pull/951

jacksongoode:
This would be a huge update :partying_face:

robinvd (approved):

capnfabs:
@robinvd thanks for testing! 

Just pushed a commit to use librespot libraries directly, which address @hhirtz's feedback above.

I think this is ready to go from my side, but I don't know what still needs to be done. Do you have any input / feedback about next steps?

robinvd:
We now just need 2 maintainers to approve the pr

robinvd (approved):

hhirtz (approved):

capnfabs:
Hmm, still a bunch of lint checks / CI Failures; I'll see if I can work through them this weekend.

capnfabs:
Alright, after a bunch of messing around, I've _finally_ got all the checks passing 😅

There's one change I was a little suspicious about -- changing the capitalisation of TV, AVR, and STB in https://github.com/Spotifyd/spotifyd/pull/977/commits/f86275a67d8d2e90d50e350fc8323b8609ea34b8 to make clippy happy. I tested that with a config file on my local machine, and the values still deserialize correctly.

MDeiml:
Something I noticed when I worked with your changes: I think you can make the `rspotfy` dependency optional and only use it with the `dbus_mpris` feature.

SimonPersson (approved):

SimonPersson:
Thank you for this work! This looks pretty good to me. A few TODOs left, but I agree with @robinvd, they can be dealt with later. I'll go ahead and merge.

SimonPersson:
Ouch, I only reviewed the code, and missed the 13 commits. In retrospect, we probably should have squashed those before merging... :)

Issue #959, opened 2021-05-29
Track ID not changing with on_song_change_hook 


chopnico:
**Description**
The Track ID doesn't seem to change when running on_song_hook_change

**To Reproduce**
Simply changing the song and looking at the logs.

**Expected behavior**
I would except the track ID would match the song that triggered the onevent

**Logs**
```
May 29 17:12:04 hildegard spotifyd[30784]: Running "/home/<omit>/.config/spotifyd/command -id <omit> -secret <omit>" using "/bin/bash" with environment variables {"DURATION_MS": "352000", "PLAYER_EVENT": "play", "POSITION_MS": "0", "TRACK_ID": "3gOJUhQe0AHfLR9Wby6c4J", "PLAY_REQUEST_ID": "23"}
May 29 17:12:05 hildegard spotifyd[30784]: Running "/home/<omit>/.config/spotifyd/command -id <omit> -secret <omit>" using "/bin/bash" with environment variables {"TRACK_ID": "7euBZraCkFV5bbct6mn2KP", "PLAYER_EVENT": "change", "OLD_TRACK_ID": "3gOJUhQe0AHfLR9Wby6c4J"}
May 29 17:12:05 hildegard spotifyd[30784]: Loading <Beyond The Lost Sky> with Spotify URI <spotify:track:4Rvcs1jHrpTzBzpItH3wuA>                
May 29 17:12:05 hildegard spotifyd[30784]: <Beyond The Lost Sky> (379946 ms) loaded                                                             
May 29 17:12:44 hildegard spotifyd[30784]: Running "/home/<omit>/.config/spotifyd/command -id <omit> -secret <omit>" using "/bin/bash" with environment variables {"PLAY_REQUEST_ID": "24", "PLAYER_EVENT": "load", "POSITION_MS": "0", "TRACK_ID": "7euBZraCkFV5bbct6mn2KP"}
May 29 17:12:45 hildegard spotifyd[30784]: Running "/home/<omit>/.config/spotifyd/command -id <omit> -secret <omit>" using "/bin/bash" with environment variables {"PLAYER_EVENT": "play", "PLAY_REQUEST_ID": "24", "TRACK_ID": "7euBZraCkFV5bbct6mn2KP", "POSITION_MS": "0", "DURATION_MS": "319866"}
[...]

Issue #913, opened 2021-03-09
"play" event is delayed until last 40 or so seconds of the track


panabar:
**Description**
I'm using on_song_change_hook to update my status bar. When spotifyd triggers a "play" PLAYER_EVENT, I request the song name via playerctl and update my statusbar.

In version 0.3.2 "play" PLAYER_EVENT frequently gets triggered in the last few seconds (40 or so just before the "load" event) of the current song. The last output to stdout when the song starts playing is usually "change" and sometimes "endoftrack". 

I have also checked if my script is the culprit by changing the on_song_change_hook to the following.
```
#!/bin/sh
echo "exiting..."
exit 0
```

Also, with this simple script the "play" even was delayed until just before the "load" event.

I am grateful and spotifyd makes me happy. Thank you. 

**To Reproduce**

**Compilation flags** 
- [ ] dbus_mpris
- [x] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [x] pulseaudio_backend
- [x] rodio_backend

**Versions (please complete the following information):**

- OS: Linux artix49 5.11.4-zen1-1-zen #1 ZEN SMP PREEMPT Sun, 07 Mar 2021 19:53:46 +0000 x86_64 GNU/Linux

- Spotifyd: 0.3.2
- cargo: NA

Barbarossa93:
It's strange, with my script, it gets triggered a few times in the beginning of the track (usually 2-3 times) and then again near the end. If I do something simple like a `notify-send test` Then it happens once in the beginning and once near the end. Is there some sort of internal timeout with `on_song_change_hook` so that if the command doesn't finish early enough it executes it again?

PR #954, opened 2021-05-15
add configuration option for bus type

eladyn:
This adds a new configuration option to configure which bus should be used by spotifyd. It addresses situations like #944 or one I personally found me in, where access to the MPRIS interface on headless systems would really help a lot.

Since I'm quite new to the codebase as well as the Rust ecosystem in general, I'm not sure if everything I changed / added was a good idea. Therefore, I'm always happy to take improvement proposals and adapt the PR accordingly. :grinning:

PR #1047, opened 2022-01-26
replace hook processes with tokio::process

eladyn:
As outlined in https://github.com/Spotifyd/spotifyd/issues/959#issuecomment-1016631025, handling too many events in a row can lead to the channel not being polled enough and the events then stacking up in the channel.

This fixes the behavior by implementing the hook process management asynchronously. ~~After the process has finished, the output is written with blocking `io::stdout()` to stdout (as before). I'll probably reimplement this in a non-blocking manner, thus still a "draft".~~

I would be happy, if someone else could already test the PR, for me it's working fine!

related: #959, #913

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ serde = { version = "1.0.115", features = ["derive"] }
 sha-1 = "0.9"
 structopt = "0.3.17"
 syslog = "4"
-tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread"] }
+tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread", "process"] }
 tokio-compat = { version = "0.1.6", features = ["rt-current-thread"] }
 tokio-compat-02 = "0.2.0"
 tokio-stream = "0.1.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ serde = { version = "1.0.115", features = ["derive"] }
 sha-1 = "0.9"
 structopt = "0.3.17"
 syslog = "4"
-tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread", "process"] }
+tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread", "process", "io-std"] }
 tokio-compat = { version = "0.1.6", features = ["rt-current-thread"] }
 tokio-compat-02 = "0.2.0"
 tokio-stream = "0.1.7"
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -89,7 +89,7 @@ pub(crate) struct MainLoopState {
     pub(crate) autoplay: bool,
     pub(crate) volume_ctrl: VolumeCtrl,
     pub(crate) initial_volume: Option<u16>,
-    pub(crate) running_event_program: Option<Child>,
+    pub(crate) running_event_program: Option<Pin<Box<Child>>>,
     pub(crate) shell: String,
     pub(crate) device_type: DeviceType,
     pub(crate) use_mpris: bool,
@@ -117,22 +117,35 @@ impl Future for MainLoopState {
             }
 
             if let Some(mut child) = self.running_event_program.take() {
-                match child.try_wait() {
-                    // Still running...
-                    Ok(None) => self.running_event_program = Some(child),
-                    // Exited with error...
-                    Err(e) => error!("{}", e),
-                    // Exited without error...
-                    Ok(Some(_)) => (),
+                // check if child has already exited
+                if let Poll::Ready(result) = child.as_mut().poll(cx) {
+                    match result {
+                        // Exited without error...
+                        Ok(_) => (),
+                        // Exited with error...
+                        Err(e) => error!("{}", e),
+                    }
+                } else {
+                    // drop the Box that holds a reference to our child
+                    self.running_event_program = Some(child);
                 }
             }
+
             if self.running_event_program.is_none() {
                 if let Some(ref mut player_event_channel) = self.spotifyd_state.player_event_channel
                 {
                     if let Poll::Ready(Some(event)) = player_event_channel.poll_next_unpin(cx) {
                         if let Some(ref cmd) = self.spotifyd_state.player_event_program {
                             match spawn_program_on_event(&self.shell, cmd, event) {
-                                Ok(child) => self.running_event_program = Some(child),
+                                Ok(child) => {
+                                    self.running_event_program = Some({
+                                        let mut child = Box::pin(child);
+                                        // We poll the child once, so the waker is awoken once the process finishes.
+                                        // Polling on a Poll::Ready(...) is allowed in this case.
+                                        let _ = child.as_mut().poll(cx);
+                                        child
+                                    })
+                                }
                                 Err(e) => error!("{}", e),
                             }
                         }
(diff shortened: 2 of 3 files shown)


temax6:
Thanks very much, I needed the endoftrack PLAYER_EVENT for a script. This fixes the player event hook for me.

robinvd:
nice! looks good but will have to test later

eladyn:
Regarding this PR, I was thinking wether a queue based model wouldn't make more sense. Currently, if the hook takes incredibly long, the librespot events will still stack up in the channel and there will still be an increasing delay / offset, since we only run a single process at a time. (Of course, if the script takes that long, it should definitely be fixed.)

However, especially in regard to #1025, where the DBUS implementation depends on the hook programs running fast enough (since it uses the same logic as the hooks), it might be preferable to maintain a separate (bounded) queue for the hook programs, on top of which new events are added and that would then be "consumed" by the hook programs one after the other.

This would ensure that the DBUS events are sent as fast as possible *and* that hook programs can take their time to process the events properly.

All this wouldn't change the fact that we'd still need `tokio::process` for the loop waking part, so this PR would still be relevant. 

If you think that, what I wrote above, sounds reasonable, I can try to implement something like this. (Would probably wait for #1025 being merged though to avoid too many conflicts.)

(Thanks for reviewing the PRs by the way, @robinvd)

robinvd:
if i read it correctly, you mean running the hook asynchronously? so dont block on the hook. That does sound good, but does make a more complicated

eladyn:
Sorry, I could have phrased my proposal more clearly.

What I actually meant to say, is:

- when new events arrive, push them in a `VecDeque`; if the `VecDeque` exceeds a certain length (e.g. 5), discard the oldest events
- when no process is running (or the last one has finished), remove the oldest event from the `VecDeque` and start a new hook with this event

This way, we would ensure that:

- even if the hook program does take a bit longer, no events are discarded
- if the hook program is taking way too long, the "event delay" never grows too big
- MPRIS signals are sent as fast as possible
- only one process is running at a time (prevents possible race conditions in the hook, keeps the logic simple and close to the current implementation)

I hope that this was somewhat understandable. :grinning:

robinvd:
tested works good!

robinvd (approved):

NNEU-1:
> Regarding this PR, I was thinking wether a queue based model wouldn't make more sense. Currently, if the hook takes incredibly long, the librespot events will still stack up in the channel and there will still be an increasing delay / offset, since we only run a single process at a time. (Of course, if the script takes that long, it should definitely be fixed.)
> 
> However, especially in regard to #1025, where the DBUS implementation depends on the hook programs running fast enough (since it uses the same logic as the hooks), it might be preferable to maintain a separate (bounded) queue for the hook programs, on top of which new events are added and that would then be "consumed" by the hook programs one after the other.
> 
> This would ensure that the DBUS events are sent as fast as possible _and_ that hook programs can take their time to process the events properly.
> 
> All this wouldn't change the fact that we'd still need `tokio::process` for the loop waking part, so this PR would still be relevant.
> 
> If you think that, what I wrote above, sounds reasonable, I can try to implement something like this. (Would probably wait for #1025 being merged though to avoid too many conflicts.)
> 
> (Thanks for reviewing the PRs by the way, @robinvd)

Thanks for this PR !
I have noticed that (using #1025) DBus properties changed messages seem to stop after Spotifyd ran for some time. 

It sounds like this PR would address this issue. 

Unfortunately it seems that Spotifyd is getting neglected a bit, with the best PRs taking ages to get merged.

Since this PR and #1025 are so related, could you please consider implementing this PR in your fork and then making a new unified PR?

That might help getting these two PRs merged more quickly.

Thanks !

eladyn:
> Thanks for this PR ! I have noticed that (using #1025) DBus properties changed messages seem to stop after Spotifyd ran for some time.

Yeah, this also happened to me once, but since I'm currently not using that branch daily, I did not yet have time to investigate and did not assume that this was related to the PR. If it happens to me again, I might investigate a bit.

> It sounds like this PR would address this issue.

I don't think so, since this PR only helps if you have `on_song_change_hook` enabled and even then, it should not stop firing the events, it would just have an increasing offset in the events that are fired. The two PRs are definitely related as #1025 does rely on the events being polled in the same loop that is also used for the hook.

> Since this PR and #1025 are so related, could you please consider implementing this PR in your fork and then making a new unified PR?
> 
> That might help getting these two PRs merged more quickly.

I actually was considering merging some of the recent PRs in a branch to be able to use them all on my machine. If you are interested in using this, I could publish that in my fork. However, I don't think that creating one big PR really helps with getting them merged more quickly, since this does only mean more difficulties reviewing them for the maintainers.

NNEU-1:
`If you are interested in using this, I could publish that in my fork. `

that would be great, thanks!

I tried merging them myself but didn’t get very far.

I’ll test it and let you know if I come across any problems

eladyn:
@NNEU-1 FYI, i pushed the branch to https://github.com/eladyn/spotifyd/tree/pr_collection. It contains this PR, #1025, #1039, #1059 and #954.

NNEU-1:
Thanks ! I have tested this branch, unfortunately the issue is still present. It seems like it takes longer for it to fail, but it eventually does.

eladyn:
@NNEU-1 It happened to me again today and I investigated a bit:

It seems that this has something to do with token expiration for the Spotify API.

```
spotifyd[75027]: <Song> (220000 ms) loaded
spotifyd[75027]: loop <- this means that the dbus loop is receiving and handling the player events
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: token is expired
spotifyd[75027]: starting token request
spotifyd[75027]: token is expired
spotifyd[75027]: token result: Pending
spotifyd[75027]: token is expired
spotifyd[75027]: token result: Pending
spotifyd[75027]: token is expired
spotifyd[75027]: token result: Ready(Ok(Token { ... }))
spotifyd[75027]: loop
spotifyd[75027]: Loading <Song> with Spotify URI <spotify:track:0D93F938DEADNB938Dxnd2>
spotifyd[75027]: <Song> (220000 ms) loaded
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: Loading <Song> with Spotify URI <spotify:track:0D93F938DEADNB938Dxnd2>
spotifyd[75027]: <Song> (220000 ms) loaded
spotifyd[75027]: loop
# manually requesting the Metadata over dbus leads to the following
spotifyd[75027]: Couldn't fetch metadata from spotify: Err(Unauthorized)
spotifyd[75027]: loop
spotifyd[75027]: loop
spotifyd[75027]: loop
```

So the DbusServer is properly requesting a new token after the old one has expired but apparently the other things don't really take notice of this (?) and continue using the invalid token. Strangely, the DBus loop doesn't print the `Couldn't fetch metadata from spotify`. I might dig further into this, but this is almost definitely not a problem with this PR or #1025 (I actually noticed previously - even before #1025 - that the DBus API stopped returning sensible results).

NNEU-1:
Ahh i forgot that Spotifyd still communicates with the Spotify api separately from librespot.
Good to know where the error lies - would be great if it were fixed.

eladyn:
This has been merged together with #1059. :tada:

Issue #990, opened 2021-08-24
don't bind to all interfaces (0.0.0.0) by default


ayushnix:
I recently noticed that `spotifyd` binds to `0.0.0.0`. Does it need to do that? Can I configure it to listen only on localhost or on my private IP within my LAN? If not, shouldn't there be an option to not listen on all interfaces by default?

H-M-H:
Looks like this is a `librespot` problem: https://github.com/librespot-org/librespot/blob/v0.1.5/connect/src/discovery.rs#L234

Zeroconf discovery is hardcoded to bind to `0.0.0.0`.

ayushnix:
> Looks like this is a `librespot` problem

Is spotifyd somehow dependent on librespot?

H-M-H:
Yeah it is: https://github.com/Spotifyd/spotifyd/blob/f4909621da43636185f2d2ffe4392178b8d654b6/Cargo.toml#L34

ayushnix:
Thanks, I'll take this issue to the librespot repository.

ayushnix:
Librespot does provide an option to disable zeroconf but spotifyd doesn't.

https://github.com/librespot-org/librespot/blob/dev/src/main.rs#L374

ayushnix:
https://drewdevault.com/2021/10/26/stalebot.html

robinvd:
feel free to PR to remove stale bot :) with the provided link

ayushnix:
> feel free to PR to remove stale bot :) with the provided link

[Done](https://github.com/Spotifyd/spotifyd/pull/1061).

aldenjenkins:
I would like to use a specific bind ip. With multiple IP addresses on one machine, this is a critical need.

PR #1061, opened 2022-02-24
remove the stale bot

ayushnix:
The [stale bot](https://github.com/marketplace/stale) is considered [harmful](https://drewdevault.com/2021/10/26/stalebot.html) and for pretty good reasons in my opinion. It masks and hides [legitimate issues](https://github.com/Spotifyd/spotifyd/issues/990) just because they may not receive attention for a few months. Closing issues because they're old doesn't make them go away, it just hides them.


JojiiOfficial:
> Closing issues because they're old doesn't make them go away, it just hides them

I don't agree. When maintaining a big project, you'll encounter a lot of "Issues" which ain't actually issues of the project itself but rather issues with the OS or the user doing something wrong. There were a lot of issues which seemed not to be actual issues and nobody cared anyways so we added stale. In general its a good idea but I think stale should ignore issues which have been reopened once and it definitely should reopen an issue again when a new comment has been made after closing it.

In general I agree with removing stale since it has mostly only caused trouble and since I think as maintainer user opinions should be strongly considered as well.

JojiiOfficial (approved):

ayushnix:
> When maintaining a big project, you'll encounter a lot of "Issues" which ain't actually issues of the project itself but rather issues with the OS or the user doing something wrong.

I can empathize with this situation but if it ends up masking legit issues along with illegitimate issues, can we really call the bot beneficial?

Of course, this is just my humble opinion.

robinvd (approved):

PR #1067, opened 2022-03-10
improvements to the documentation

### LINKED ISSUE #1031: https://spotifyd.github.io/spotifyd/config/File.html conf file issue
Hello,
on the page https://spotifyd.github.io/spotifyd/installation/Raspberry-Pi.html
I found out that configuration file ~/.config/spotifyd/spotifyd.conf example is corrupted.
Line with :

> device = alsa_audio_device # Given by `aplay -L`  

should have value in quotes

> device = "alsa_audio_device" # Given by `aplay -L`

It invalidates TOML so spotifyd will throw error:
Error:
   0: could not load the config file
   1: invalid TOML value, did you mean to use a quoted string? at line 5 column 10

Location:
   /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/result.rs:1914

note:
   the config format should be valid TOML
   we recently changed the config format, see https://github.com/Spotifyd/spotifyd/issues/765

eladyn:
This PR adds the new page that was added in #1030 to the summary, so that it also appears in the sidebar.

It also removes the configuration sections from individual installation pages and replaces them with links to the configuration section. This should make sure that there is more or less a single source of truth regarding the config file.

The "Running as a Service" section was previously a draft page, which was replaced with a proper page shortly summarizing the contents of this chapter.

Closes #1052, since it removes the relevant section.
Fixes #1031 (now points to a section with the correct example config)


docs/src/installation/FreeBSD.md
```diff
@@ -14,24 +14,9 @@ This guide will help you to install `spotifyd` on FreeBSD and have it always run
...
-
-
-Input the following data. Change the **username**, **password**, **device_name**, and the **bitrate**.
-
-```toml
-[global]
-username = "USER"
-password = "PASS"
-backend = "portaudio"
-device = "/dev/dsp"
```
robinvd, line 27: maybe keep these 2 settings? i think the are required for freebsd but not sure

eladyn: Should've read what I was removing more carefully :sweat_smile:

I just fired up a FreeBSD-VM and tested spotifyd there without any configuration (except for username and password). It did indeed not work since spotifyd is compiled there [with only portaudio enabled](https://codeberg.org/FreeBSD/freebsd-ports/src/commit/775d857db4e7eb90ca78745da1e7f15fc8f2e50b/audio/spotifyd/Makefile#L419), but the fallback in spotifyd [is always alsa](https://github.com/Spotifyd/spotifyd/blob/ae6dac7a54f899316674ba57ce4a0f9890cd2b1c/src/config.rs#L604-L608) (which should probably be fixed to be more feature-agnostic). However, after specifying `backend = "portaudio"` everything worked fine, so the `device` is apparently picked correctly by default.

So I guess, I'll add a mention about the backend and we can maybe remove that after the fallback has been fixed.

eladyn: I readded the backend thing, do you think that we can omit the device part or should I mention that too?

@rosorio Maybe you can have a look at that paragraph as well, since it was you who originally added that section?

debdrup: I can confirm that these are needed, as it's what I'm using on my daily driver T480s running FreeBSD and Wayland+Sway.

Additionally [portaudio is the default for the FreeBSD port/package](https://www.freshports.org/audio/spotifyd#config), because portaudio maintains the same OSS compatibility that FreeBSD still has (and has had since Linux had OSS compatibility).

debdrup:
The FreeBSD port/package [comes with the rc.d service file](https://cgit.freebsd.org/ports/tree/audio/spotifyd/files/spotifyd.in) necessary to start spotifyd as a service, so if you want to add instructions for how to run as a service on FreeBSD, it's as simple as:

1: Edit config file

2: `service spotifyd enable`

3: `service spotifyd start`

docs/src/installation/FreeBSD.md
eladyn: @debdrup Thank you for the feedback! I will also add the device value back to the documentation.

Would you mind sharing what `spotifyd` does / why it doesn't work, when running without the `device` parameter? (Potentially in a new issue so that it is easier to keep track of that.) Because the ideal behaviour is of course that `spotifyd` just starts on most setups without any configuration needed. For the `backend` config, I'm already quite sure what is causing this, but the `device` config is not required in my VM, so I can't test this. Thank you!

eladyn:
@debdrup Looks like a good suggestion! As I have only little experience with FreeBSD, I'm not really sure how to fit that in here correctly (especially since there seems to be already [something similar in the docs](https://github.com/Spotifyd/spotifyd/blob/ae6dac7a54f899316674ba57ce4a0f9890cd2b1c/docs/src/installation/FreeBSD.md?plain=1#L36-L48)). Feel free to open a separate PR for this (or otherwise suggest changes directly to this one)!

docs/src/installation/FreeBSD.md
rosorio: > maybe keep these 2 settings? i think the are required for freebsd but not sure

Backend and device settings are required for FreeBSD environment.
Currently spotifyd is build to only support portaudio backend.

debdrup: I think I phrased myself poorly, because the device parameter isn't required on FreeBSD, unless you want to play on a different pair of speakers (ie. `/dev/pcm1` if `/dev/pcm0` is the default).

debdrup:
> @debdrup Looks like a good suggestion! As I have only little experience with FreeBSD, I'm not really sure how to fit that in here correctly (especially since there seems to be already [something similar in the docs](https://github.com/Spotifyd/spotifyd/blob/ae6dac7a54f899316674ba57ce4a0f9890cd2b1c/docs/src/installation/FreeBSD.md?plain=1#L36-L48)). Feel free to open a separate PR for this (or otherwise suggest changes directly to this one)!

I'll be honest, I hadn't seen those docs. :blush: 

`service <entry> enable` is just [a wrapper around sysrc](https://cgit.freebsd.org/src/blame/libexec/rc/rc.subr#n1074), so it's perfectly correct to use that over the other.

docs/src/installation/FreeBSD.md
eladyn: @rosorio Just to make sure: If you omit the `device` parameter from the config (but leave the `backend = "portaudio"`), does it really stop working and, if so, could you open an issue with details about potential errors? Otherwise, I would leave it as it is now (see [here](https://github.com/Spotifyd/spotifyd/blob/f075168162dd221cf79bc1cbf940215684e20c69/docs/src/installation/FreeBSD.md?plain=1#L19) for the current version, which just mentions the `backend` thing). Thank you both!

rosorio: > I readded the backend thing, do you think that we can omit the device part or should I mention that too?
> 
> @rosorio Maybe you can have a look at that paragraph as well, since it was you who originally added that section?



> I readded the backend thing, do you think that we can omit the device part or should I mention that too?
> 
> @rosorio Maybe you can have a look at that paragraph as well, since it was you who originally added that section?

Thanks @eladyn, you have my +1 to push your changes.

robinvd (approved):

SimonPersson:
Thanks!

PR #1092, opened 2022-06-28
Expose librespots support for limiting the cache size 

### LINKED ISSUE #1066: limit cache size config
Thanks for spotifyd. I have to clear my cache manually every two weeks, because the cache on my raspberry2 is full and the daemon — and of coarse other processes as well — stucks and doesn't play any songs.
Limit  the maximum cache size possibility in config file would be nice.
Delete the oldest when max size is reached would be the smartest way, if it is easy to implement. But clear cache completely is also a valid solution for me and the easiest way....
THX

weiznich:
Fixes #1066

### DIFF
--- a/docs/src/config/File.md
+++ b/docs/src/config/File.md
@@ -72,6 +72,10 @@ bitrate = 160
 # shell placeholders like $HOME or ~ don't work!
 cache_path = "cache_directory"
 
+# The maximal size of the cache directory in bytes
+# The example value corresponds to ~ 1GB
+max_cache_size = 1000000000
+
 # If set to true, audio data does NOT get cached.
 no_audio_cache = true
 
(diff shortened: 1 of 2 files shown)


eladyn (changes requested):
Thanks for the PR!

There are still some places, where you'd need to add the new flag as well:

- [merging of the different configuration sources](https://github.com/Spotifyd/spotifyd/blob/7cd64ff1f39c457fa92e8d15649fab02ddff013d/src/config.rs#L509-L529), otherwise the value can't be set in the config file
- [struct debugging](https://github.com/Spotifyd/spotifyd/blob/7cd64ff1f39c457fa92e8d15649fab02ddff013d/src/config.rs#L442-L466), so that the value is logged with `--verbose`
- [docs](https://github.com/Spotifyd/spotifyd/blob/7cd64ff1f39c457fa92e8d15649fab02ddff013d/docs/src/config/File.md?plain=1#L9-L101), if possible, with some documenting lines

eladyn (approved):
LGTM and works as it should!

weiznich:
@slondr Any chance on merging this?

robinvd (approved):

slondr (approved):
I haven't tested this out in a real setup but the code changes look fine

PR #1100, opened 2022-08-09
MPRIS Device Name

klay2000:
Added a compilation option to append device names to the end of dbus paths, this allows multiple Spotifyd daemons to be running each with a unique dbus object for MPRIS as long as they have unique device names.

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -166,7 +166,10 @@ async fn create_dbus_server(
 
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
-    let path = format!("org.mpris.MediaPlayer2.spotifyd.instance{b}", b = std::process::id());
+    let path = format!(
+        "org.mpris.MediaPlayer2.spotifyd.instance{b}",
+        b = std::process::id()
+    );
     conn.request_name(path, true, true, true)
         .await
         .expect("Failed to register dbus player name");
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -168,6 +168,7 @@ async fn create_dbus_server(
         "org.mpris.MediaPlayer2.spotifyd.instance{}",
         std::process::id()
     );
+    
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
     conn.request_name(path, true, true, true)
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -164,12 +164,12 @@ async fn create_dbus_server(
         panic!("Lost connection to D-Bus: {}", err);
     });
 
-    // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
-    // later. We should instead properly release the name when the session ends.
     let path = format!(
         "org.mpris.MediaPlayer2.spotifyd.instance{}",
         std::process::id()
     );
+    // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
+    // later. We should instead properly release the name when the session ends.
     conn.request_name(path, true, true, true)
         .await
         .expect("Failed to register dbus player name");
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -168,7 +168,7 @@ async fn create_dbus_server(
         "org.mpris.MediaPlayer2.spotifyd.instance{}",
         std::process::id()
     );
-    
+
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
     conn.request_name(path, true, true, true)
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -167,8 +167,8 @@ async fn create_dbus_server(
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
     let path = format!(
-        "org.mpris.MediaPlayer2.spotifyd.instance{b}",
-        b = std::process::id()
+        "org.mpris.MediaPlayer2.spotifyd.instance{}",
+        std::process::id()
     );
     conn.request_name(path, true, true, true)
         .await
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -166,7 +166,8 @@ async fn create_dbus_server(
 
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
-    conn.request_name("org.mpris.MediaPlayer2.spotifyd", true, true, true)
+    let path = format!("org.mpris.MediaPlayer2.spotifyd.instance{b}", b = std::process::id());
+    conn.request_name(path, true, true, true)
         .await
         .expect("Failed to register dbus player name");
 


eladyn:
Hi and thanks for the contribution!

While your approach certainly works, I think that it'd be much nicer, if this setting either were the default or customizable at runtime via some configuration option. The reason for this is that most people probably download a prebuilt `spotifyd` binary rather than building it themselves, so they never get to choose the dbus name behaviour.

Happy to hear your thoughts on that!

klay2000:
A configuration option sounds like a good idea, I'll give that a try.

src/dbus_mpris.rs
```diff
@@ -153,7 +159,16 @@ async fn create_dbus_server(
...
 
-    conn.request_name("org.mpris.MediaPlayer2.spotifyd", false, true, true)
+    let path:String;
+    
+    if unique_mpris {
+        path = format!("{a}_{b}", a=DBUS_PATH, b=&device_name);
+    }
+    else{
+        path = DBUS_PATH.to_string();
+    }
```
eladyn, line 169: We can avoid allocations here using [`Cow`s](https://doc.rust-lang.org/std/borrow/enum.Cow.html). (Would of course be better to import `std::borrow::Cow` below.)

```suggestion
    let path: std::borrow::Cow<str> = if unique_mpris {
        format!("{DBUS_PATH}_{device_name}").into()
    } else {
        DBUS_PATH.into()
    };
```

src/config.rs
```diff
@@ -288,6 +288,15 @@ pub struct SharedConfigValues {
     #[cfg_attr(not(feature = "dbus_mpris"), structopt(skip), serde(skip))]
     use_mpris: Option<bool>,
 
+    #[cfg_attr(
+        feature = "dbus_mpris",
+        structopt(long),
+        serde(alias = "unique-mpris", default)
+    )]
+    #[cfg_attr(not(feature = "dbus_mpris"), structopt(skip), serde(skip))]
+    unique_mpris: Option<bool>,
```
eladyn, line 297: Would be good to add some short documentation string above that will then appear in `spotifyd --help`.

eladyn (changes requested):
Generally looks good to me now. You'll need to run `cargo fmt` before this is ready, to keep coding style consistent. 

In the docs, there's also [the sample config](https://github.com/Spotifyd/spotifyd/blob/993336f74ec89cb6cad23dd009251e70548761b6/docs/src/config/File.md?plain=1#L9-L105), where this new option should be added as well, accompanied by some short explanation.

docs/src/config/File.md
```diff
@@ -34,6 +34,10 @@ use_keyring = true
 # then set this to false to avoid binding errors
 use_mpris = true
 
+# If set to true, the device name is appended to the debus path for MPRIS,
+# this allows multiple instances of `spotifyd` to run at once with MPRIS on.
+unique
```
eladyn, line 39:
```suggestion
# If set to true, the device name is appended to the dbus path for MPRIS,
# this allows multiple instances of `spotifyd` to run at once with MPRIS on.
unique_mpris = false
```

eladyn (approved):
One last comment. Apart from that, I think this is ready. Thank you! Now, we'll just have to wait for the maintainers.

docs/src/config/File.md
klay2000: oh that's funny, I actually had exactly this and must've forgotten to hit save

JasonLG1979:
The spec says that apps that allow multiple instances should follow [this convention](https://specifications.freedesktop.org/mpris-spec/latest/#Bus-Name-Policy).

So if the name is "spotifyd" the bus name should be:

`org.mpris.MediaPlayer2.spotifyd.instance{deviceName}`

Where `deviceName` is the sanitized device name. [Bus names can only contain ASCII characters (but no `-`)](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus)

JasonLG1979:
Really the easiest way is just to follow the spec and make the bus name `org.mpris.MediaPlayer2.spotifyd.instance{pid}` all of the time and not event use the `DeviceName` at all.

eladyn:
> The spec says that apps that allow multiple instances should follow [this convention](https://specifications.freedesktop.org/mpris-spec/latest/#Bus-Name-Policy).

Thanks for chiming in! This is indeed something that should be considered.

> So if the name is "spotifyd" the bus name should be:
> 
> `org.mpris.MediaPlayer2.spotifyd.instance{deviceName}`
> 
> Where `deviceName` is the sanitized device name. [Bus names can only contain ASCII characters (but no `-`)](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus)

If the `deviceName` doesn't start with a number, we could omit the `instance` part, but I don't think that is guaranteed. The ASCII sanitization is indeed missing (even in the current implementation).

> Really the easiest way is just to follow the spec and make the bus name `org.mpris.MediaPlayer2.spotifyd.instance{pid}` all of the time and not even use the `DeviceName` at all.

I can imagine that some use cases might benefit from the explicit naming with `DeviceName`, since it makes the bus names reproducible between restarts and can be hard coded in a custom client that targets specifically one instance. I personally don't have a use case for this flag, so I don't know in which situations people would be using it. Might as well be that the pid version is just sufficient for those who need it. Probably something that @klay2000 can answer best.

JasonLG1979:
> I don't know in which situations people would be using it. Might as well be that the pid version is just sufficient for those who need it. Probably something that @klay2000 can answer best.

I'm not sure either? If it's just for the sake of uniqueness purely for allow for multiple instances I would follow the specs suggestion, and really that could just be the default behavior all of the time.

On the other hand if it's to let a user know what the instance is in any sort of UI you'd want to set the [Identity](https://specifications.freedesktop.org/mpris-spec/latest/Media_Player.html#Property:Identity) to something like `Spotifyd {DeviceName}` Identity can be any arbitrary string, it's not bound by the bus name rules, and that's what would displayed to a user not the bus name. You may already do this idk?

klay2000:
Reading the spec, it actually only says that it must have a dot and a unique identifier appended to it's bus name, the ```instance{pid}``` thing is just an example so I *think* something like ```org.mpris.MediaPlayer2.spotifyd.{DeviceName}``` is sufficient to conform to the spec.

As far as ASCII sanitization goes I definitely forgot about that one, I'll make these changes and push it up.

JasonLG1979:
Do you need it to be the device name for any reason in particular? If it's just for the sake of having a unique name I strongly suggest using what the example suggests as it is the convention that all other players follow. It's also useful to clients because if it's the pid it can be used for other things.

JasonLG1979:
Pids don't change over the course of a program's life. You're prob getting a pid of a child process/thread of Spotifyd mixed up with the main process/thread.

JasonLG1979:
Not to mention that 2 equally valid device name could be sanitized into the same name thus breaking the uniqueness guarantee.

klay2000:
Exactly, the parent process is dying and leaving several child processes.

klay2000:
Alright, I'll change it to the PID.

JasonLG1979:
My guess would be that you want the pid of the Tokio runtime.

JasonLG1979:
If you try to grab a pid before you're in the runtime it may be wrong?

JasonLG1979:
> I suppose either is useful but I've been debugging something that uses this feature and I've been running into issues with Spotifyd's PID not being the same as the one it has on creation when I try to kill it's process, so at the moment it seems like it's not going to be as easy to make use of.

I sprinkled some `println!("PID in 'whatever': {}", std::process::id());` all around the current dev version of librespot and they all printed the PID.

My guess is that if you're using systemd (or some other way to daemonize Spotifyd) is that you're not getting the actual PID but the PID of the task that spawns Spotifyd?

JasonLG1979:
This whole PR could be as simple as:
```rust

    let busname = format!(
        "org.mpris.MediaPlayer2.spotifyd.instance{}",
        std::process::id()
    );
    
    conn.request_name(&busname, true, true, true)
        .await
        .expect("Failed to register dbus player name");
```

src/dbus_mpris.rs
```diff
@@ -161,15 +161,7 @@ async fn create_dbus_server(
...
-        format!(
-            "{a}.{b}",
-            a = DBUS_PATH,
-            b = device_name
-                .chars()
-                .filter(|c| c.is_ascii())
-                .collect::<String>()
-        )
-        .into()
+        format!("{a}.instance{b}", a = DBUS_PATH, b = std::process::id()).into()
```
JasonLG1979, line 164: This could totally just be the default behavior. There's no need to have config option. Single instance users lose nothing and you gain multi-instance support out of the box. It's a -0 +1 situation.

JasonLG1979: How?

klay2000: Any software that interfaces via dbus would no longer work until they make that change.

JasonLG1979: `dbus-send --print-reply --dest=org.freedesktop.DBus  /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep org.mpris.MediaPlayer2.spotifyd.instance`

JasonLG1979: > Any software that interfaces via dbus would no longer work until they make that change.

No one should be writing software designed to work with just one specific MPRIS interface. That's broken by design in the 1st place.

JasonLG1979: By "software" ofc you mean hacky bash scripts,lol!!!

JasonLG1979: I plan on adding a command API to librespot similar to [librespot-java's](https://github.com/librespot-org/librespot-java/tree/dev/api) and I've got a pending PR that greatly expands the usefulness of the events (which are trivial to UDP multicast with a few lines of python) so abusing MPRIS in that way will not be necessary in the future.

JasonLG1979: Another option is to check to see if `org.mpris.MediaPlayer2.spotifyd` already exists, if it doesn't just use that name if not use `org.mpris.MediaPlayer2.spotifyd.instance{pid}` That would satisfy both backwards compatibility for existing scripts and the unique name requirement. You also have to un-own the name on drop though to make sure not to orphan it.

eladyn: Thanks to both of you for the various points. I'm personally the author of such a “software” that currently relies on the MPRIS name being exactly the one that it is now. In my case, though, it should be quite trivial to adapt that to the PID behaviour, since I'm listening for new bus names anyway.

From what I read in the various issues in this repo, there are a bunch of people who are using such scripts to control their `spotifyd` instances and as such rely on the name being always the same. If you consider, however, how easy it should be to add support for the new bus names to their scripts, this breakage might be acceptable?

To do this, you could:

- use `playerctl`, that Just Works™ with the new behaviour, even if you specify `playerctl -p spotifyd ...`
- use one of the methods outlined [here](https://unix.stackexchange.com/a/46309) to get a list of all bus names and filter that with `grep "org.mpris.MediaPlayer2.spotifyd"` to get the bus name

If we decide to unconditionally change that, it might still be good to note that and possible solutions somewhere.

JasonLG1979: @eladyn you could easily just use the grep method or similar in playerctl so a user would not need to know the full busname and if it matches multiple instances the throw an an error that the user needs to be more specific.

eladyn: Yeah, what I meant to say with `playerctl` is that when I do:

```
playerctl -p spotifyd status
```
it works out of the box, even if the bus name is e.g. `org.mpris.MediaPlayer2.spotifyd.instance9928`. So there would be no change needed.

JasonLG1979 (commented):
You don't need all the b = stuff, you've only got one value and format args can be positional. The only reason to assign them if it needs to be repeated in the string.

src/dbus_mpris.rs
```diff
@@ -166,7 +166,11 @@ async fn create_dbus_server(
 
     // TODO: The first `true` allows us to replace orphaned dbus servers from previous sessions
     // later. We should instead properly release the name when the session ends.
```
eladyn, line 168: Could you move that comment below your insertion so that it stays next to the relevant line?

eladyn (approved):
LGTM, thanks for your work!

robinvd (approved):

PR #1105, opened 2022-08-19
fix clippy lints

eladyn:
This fixes a [new clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) that recently caused [checks to fail](https://github.com/Spotifyd/spotifyd/runs/7926327267?check_suite_focus=true).


SimonPersson:
Thanks!

PR #1113, opened 2022-09-08
make `Seek` and `SetPosition` functions spec compliant

### LINKED ISSUE #1112: MPRIS Seek and SetPosition implementations are incompatible with what playerctl expects
**Description**

Basic playback control with `playerctl` is possible, but triggers an error or segfault when attempting to seek to an absolute/relative position. When inspecting the D-Bus interfaces with `d-feet`, it becomes obvious that `spotifyd` offers a different interface from all other software implementing the MPRIS specification.

The [Seek](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Method:Seek) method expects a signed int64 to be able to seek backwards and forwards. Additionally, [it is currently interpreted in milliseconds](https://github.com/Spotifyd/spotifyd/blob/7146f64a819590a16de9991de74f8bc2fea9c2ac/src/dbus_mpris.rs#L268), but should be interpreted in microseconds.

The [SetPosition](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Method:SetPosition) method expects a string trackid (the rationale being "to avoid race conditions where a client tries to seek to a position when the track has already changed") and a signed int64 in microseconds. Given that the implementation looks very similar to the Seek one, it's likely it's interpreted in milliseconds instead.

**To Reproduce**

- `playerctl -p spotifyd position 10` to set absolute position to 10s
- `playerctl -p spotifyd position 10+` to seek forward by 10s
- `playerctl -p spotifyd position 10-` to seek backward by 10s

To use d-feet:

- `d-feet`
- Click on "Session Bus"
- Narrow down search results by entering "mpris"
- Select the spotifyd one
- Click on the `/org/mpris/MediaPlayer2` path
- Click on the `org.mpris.MediaPlayer2.Player` interface
- Click on the methods

**Expected behavior**

- spotifyd playback continues at 10s or 10s later or 10s earlier

**Actual behavior**

The first playerctl call segfaults:

    (playerctl:251381): GLib-CRITICAL **: 13:52:57.116: g_variant_new_object_path: assertion 'g_variant_is_object_path (object_path)' failed

The second and third fail with a D-Bus error:

    Could not execute command: GDBus.Error:org.freedesktop.DBus.Error.Failed: D-Bus argument type mismatch at position 0: expected u32, found i64

**Logs**
<details><summary>Click to show logs</summary>

No logs, duh

</details>

**Compilation flags** 
- [ ] dbus_mpris
- [ ] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [ ] pulseaudio_backend
- [ ] rodio_backend

- OS: Arch Linux
- Spotifyd: 0.3.3 (7146f64a819590a16de9991de74f8bc2fea9c2ac) https://aur.archlinux.org/packages/spotifyd-full-git
- cargo: 1.63.0

eladyn:
This makes some MPRIS methods more spec compliant. @wasamasa does this resolve all your issues?

I'm not sure, if the integer type casting is overly cautious the way it is currently.

Fixes #1112.

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -573,7 +573,7 @@ async fn create_dbus_server(
             .recv()
             .await
             .expect("Changed track channel was unexpectedly closed");
-        let mut seeked_position = None;
+        let mut seeked_position_ms = None;
 
         // Update playback state from event
         let (track_id, playback_status, player_volume) = match event {
@@ -585,7 +585,7 @@ async fn create_dbus_server(
                 position_ms,
                 ..
             } => {
-                seeked_position = Some(position_ms);
+                seeked_position_ms = Some(position_ms);
                 (Some(track_id), Some(PlaybackStatus::Playing), last_volume)
             }
             PlayerEvent::Stopped { .. } => {
@@ -672,13 +672,14 @@ async fn create_dbus_server(
         }
 
         // if position in track has changed emit a Seeked signal
-        if let Some(position) = seeked_position {
+        if let Some(position_ms) = seeked_position_ms {
             let msg = dbus::message::Message::signal(
                 &dbus::Path::new("/org/mpris/MediaPlayer2").unwrap(),
                 &dbus::strings::Interface::new("org.mpris.MediaPlayer2.Player").unwrap(),
                 &dbus::strings::Member::new("Seeked").unwrap(),
             )
-            .append1(position as i64);
+            // position should be in microseconds
+            .append1(position_ms as i64 * 1000);
             conn.send(msg).unwrap();
         }
     }
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -29,7 +29,7 @@ use rspotify::{
     prelude::*,
     AuthCodeSpotify, Token as RspotifyToken,
 };
-use std::{collections::HashMap, env, pin::Pin, sync::Arc};
+use std::{collections::HashMap, convert::TryInto, env, pin::Pin, sync::Arc};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 
 pub struct DbusServer {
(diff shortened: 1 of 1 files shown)


slondr (dismissed):

slondr:
I'd like confirmation from @wasamasa that this fixes their issue before merging this, but the code looks fine.

wasamasa:
Thanks, I've tested the PR and it removes the need for all hacks in my script used to control spotifyd. I had a bit of trouble figuring out the correct syntax for the path for `SetPosition` to work.

eladyn:
I missed one spot, where we incorrectly returned a time in milliseconds as Position, [instead of microseconds](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:Position). Now, position updates are consistent with the rest (and the spec) as well.

flying-sheep:
Looks good!

grafik

I’m seeing one last bug. Might be that it’s a KDE plasma bug though. I hope I can describe the bug correctly.

So first what works:

1. the seek position advances
2. clicking the seek bar works and resets to the correct position in the track

Now to the bug:

1. When I have the spotify desktop app open as well, I have two MPRIS players.
2. Initially, switching from “Spotify” (desktop app) to “spotifyd” resets the seek position in the MPRIS interface to 0 ← this is a bug
3. That changes after seeking in the MPRIS interface for “spotifyd”. The seek position seems healed: after switching between MPRIS interfaces and coming back to “spotifyd”, the seek position is correct.
4. But it seems that after healing this one, now the “Spotify” MPRIS interface is broken ← I feel like that implies it might be a Plasma bug

eladyn:
@flying-sheep Thanks for testing this PR! Since, what you describe happens as well with other MPRIS implementations, it seems indeed likely that this is rather a buggy client-side implementation than us implementing the server-side wrong. Maybe you can report that bug in the plasma repo, and we'll see what comes out of it.

robinvd (approved):

slondr (approved):
aight, LGTM

PR #1121, opened 2022-09-25
only use cache if username matches

eladyn:
This is a follow-up to #1019 and reorders the credential cache logic.

Previously:

- `username` and `password` are specified: use that
- cached credentials exist: use that
- else: start discovery

Now:

- cached credentials exist for given `username`: use that
- `username` and `password` are specified: use that
- else: start discovery

Also, a cache initialization failure is properly reported to the user.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -590,16 +590,11 @@ pub(crate) fn get_internal_config(config: CliConfig) -> SpotifydConfig {
     let cache = config
         .shared_config
         .cache_path
-        .map(PathBuf::from)
-        // TODO: plumb size limits, check audio_cache?
-        // TODO: rather than silently disabling cache if constructor fails, maybe we should handle the error?
-        .and_then(|path| {
-            Cache::new(
-                Some(path.clone()),
-                if audio_cache { Some(path) } else { None },
-                size_limit,
-            )
-            .ok()
+        .map(|path| Cache::new(Some(&path), audio_cache.then_some(&path), size_limit))
+        .transpose()
+        .unwrap_or_else(|e| {
+            warn!("Cache couldn't be initialized: {e}");
+            None
         });
 
     let bitrate: LSBitrate = config
--- a/src/setup.rs
+++ b/src/setup.rs
@@ -147,11 +147,16 @@ fn get_credentials(
     username: &Option<String>,
     password: &Option<String>,
 ) -> Option<Credentials> {
-    if let (Some(username), Some(password)) = (username, password) {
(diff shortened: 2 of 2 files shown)


eladyn:
@aanno, if you've got the time, I'd be happy if you could test this version and see if it fixes the problem (again :sweat_smile:).

robinvd (approved):

slondr (approved):
I'll hold off on merging until @aanno confirms, but code looks great

aanno:
Well, I tried but currently I run into problems to compile. Error is:

` ` `
   Compiling spotifyd v0.3.3 (/root/spotifyd)
error[E0658]: use of unstable library feature 'bool_to_option'
   --> src/config.rs:593:57
    |
593 |         .map(|path| Cache::new(Some(&path), audio_cache.then_some(&path), size_limit))
    |                                                         ^^^^^^^^^
    |
    = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `spotifyd` due to previous error
Error: error building at STEP "RUN set -ex;  mkdir /opt/owntone || true;   cd;   export DEBIAN_FRONTEND=noninteractive;   export FEATURES="--no-default-features --all-features --features pulseaudio_backend,portaudio_backend,alsa_backend,dbus_keyring,dbus_mpris";   ln -fs /usr/share/zoneinfo/Europe/Berlin /etc/localtime;   apt update;   apt -y install   git rustc cargo libasound2-dev libssl-dev pkg-config     portaudio19-dev libdbus-1-dev libpulse-dev libogg-dev;   echo not using git clone -b zeroconf_cache_fix https://github.com/eladyn/spotifyd.git;   echo not using git clone https://github.com/Spotifyd/spotifyd.git;   git clone -b cached_credentials_fix https://github.com/eladyn/spotifyd.git;   cd spotifyd;   cargo build $FEATURES --release;   cargo install $FEATURES --path . --locked;   cp -r target/release/spotifyd ~/.cargo /opt/owntone/;": error while running runtime: exit status 101
` ` `

eladyn:
`.then_some()` should be stable since 1.62.0, so you'd just need to upgrade your toolchain to a more recent version, and it should work. :)

aanno:
Updated my building hack at https://github.com/aanno/linux-config/tree/master/build-with-docker to build with `rustup` (instead of the `rustc` that comes with ubuntu jammy).

Yes, I tested it and it is working as expected. Great work! Thank you very much!

eladyn:
Thank you!

Issue #1123, opened 2022-10-04
Error "use of unstable library feature" while building


kristopalka:
**Description**
While compiling using cargo from latest master, error occurs: use of unstable library feature 'bool_to_option' in **src/config.rs:636:57**.
This error occurs in lastly changed lines: 00f52f983ce0d2162ea2fef677e2c9eee7c8f6ef

**To Reproduce**
1. sudo apt install rustc cargo libasound2-dev libssl-dev pkg-config
2. git clone https://github.com/Spotifyd/spotifyd.git
3. cd spotifyd
4. cargo build --release <- ERROR

**Expected behavior**
It should compile

**Logs**
<details><summary>Click to show logs</summary>

```
xxx@xxx:~/projects/spotifyd$ cargo build --release 
   Compiling spotifyd v0.3.3 (/home/xxx/spotifyd)
error[E0658]: use of unstable library feature 'bool_to_option'
   --> src/config.rs:636:57
    |
636 |         .map(|path| Cache::new(Some(&path), audio_cache.then_some(&path), size_limit))
    |                                                         ^^^^^^^^^
    |
    = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `spotifyd` due to previous error

```
</details>

**Compilation flags** 
- [ ] dbus_mpris
- [ ] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [ ] pulseaudio_backend
- [ ] rodio_backend

**Versions (please complete the following information):**

- OS: Ubuntu 20.04.5 LTS
- Spotifyd: 00f52f983ce0d2162ea2fef677e2c9eee7c8f6ef
- cargo: 1.61.0

eladyn:
Hi and thank you for the report!

The problematic `then_some` has been stabilized in Rust 1.62.0 (the latest is 1.64.0), so in order to compile this version, you'd need to use a different distribution channel (e.g. [`rustup`](https://rustup.rs/)).

---

In general, I'm not sure what a good way forward would be on this one? I don't think there are any guarantees about a minimum rust version that is supported, and, since the main distribution channel is probably through GitHub releases, it shouldn't cause much breakage either to “live on the edge”.

In this case, though, the benefit of using the feature is minor and could be easily reverted. Given that there have already been two users that stumbled upon this independently of each other, it might be worth changing that line. Alternatives would be:

```rust
[...]

PR #1127, opened 2022-10-13
Specify minimum rust version

### LINKED ISSUE #1126: Formally specifcy MSRV
We had a discussion recently about specifying a minimum Rust version. Turns out there actually is a way to put that in the codebase, so users with very old Rust versions will be told in no uncertain terms what the problem is.

slondr:
Resolves #1126

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,7 @@ jobs:
       - name: Installing Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain:
-            - "1.62"
-            - stable
+          toolchain: stable
           override: true
       - name: Installing needed Ubuntu dependencies
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,7 @@ jobs:
       - name: Installing Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain:
-            - "1.62"
-            - stable
+          toolchain: stable
           override: true
       - name: Checking out sources
         uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


robinvd (dismissed):

eladyn:
Cool, didn't know that. LGTM! I wonder if we could also pin that version [in CI](https://github.com/Spotifyd/spotifyd/blob/master/.github/workflows/ci.yml) (not CD) so that we notice, when we need to bump the MSRV.

slondr:
Good idea @eladyn. I've added that to this MR.

slondr:
Oh wait, that broke everything.

slondr:
Ok, we're good now.

.github/workflows/ci.yml
```diff
@@ -56,6 +56,8 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
+        toolchain:
+          - "1.62"
```
eladyn, line 60: That shouldn't be needed, since you aren't using `${{ matrix.toolchain }}` anywhere.

Plecra:
Fantastic, LGTM :) These are almost the exact changes I made in response to #1123 , which this also fixes (and im going to sneakily link to this now ;)) I'm going to make some small changes to CI to ensure we're still getting the most recent lints

Plecra (approved):
assuming the CI is happy, lets get this done! It'd be nice to have a landing page in the documentation for the error ubuntu users will get:

> ```sh
> error: package `spotifyd v0.3.3 (...)` cannot be built because it requires rustc 1.62 or newer,
> while the currently active rustc version is 1.61.0
> ```

Plecra:
agh, clippy 🙄 fine, will make a quick PR for those

eladyn (approved):
Thanks for picking this up!

Issue #1128, opened 2022-10-17
[Regression] Non-predictable D-Bus name


haasn:
**Description**

Commit 5270c3a3abcc5f9fec2695804e005299d5badaf7 caused a regression on my end. I can no longer control spotifyd through the D-Bus interface, which I currently do as follows:

```haskell
playerControl :: Client -> String -> X ()
playerControl c m = io . callNoReply c $
    (methodCall path iface member) { methodCallDestination = Just dest }
    where path   = objectPath_ "/org/mpris/MediaPlayer2"
          iface  = interfaceName_ "org.mpris.MediaPlayer2.Player"
          member = memberName_ m
          dest   = busName_ "org.mpris.MediaPlayer2.spotifyd"
```

As you can see, this sends a d-bus event to the `org.mpris.MediaPlayer2.spotifyd` endpoint. But the linked commit removed the ability to refer to spotifyd using a consistent name, instead I now have to traverse the D-Bus tree to find the right endpoint, in the worst case on every single command (even something as simple as changing the volume!).

What was the reasoning for this breaking change? Can it be reverted, or made configurable? Suggestions on a better way to solve my use-case?

**Compilation flags** 
- [x] dbus_mpris
- [x] dbus_keyring
- [x] alsa_backend
- [ ] portaudio_backend
- [x] pulseaudio_backend
- [ ] rodio_backend

eladyn:
Thanks for bringing this up! This is indeed not a regression, but rather a feature :sweat_smile:, and we are aware that this might be a breaking change for some users. (See the [changelog entry](https://github.com/Spotifyd/spotifyd/blob/ddfcbb2eecd3b5be94533f7374180df412659b4e/CHANGELOG.md#changed).)

In the PR that changed that (#1100), we discussed whether a unique bus name (which is, what it is now) should be the hardcoded default or if this should be configurable. (See https://github.com/Spotifyd/spotifyd/pull/1100#discussion_r965418312 for some aspects.) There, we agreed that, even if it might break some setups, it should be easy to adapt to it with some minor changes, which are also noted in the changelog.

All that aside: The new version hasn't yet been released, and there is always room for change / improvement. I don't know, how easy it would be in your case to first list all the bus names and then filter for `spotifyd`, before controlling `spotifyd`?

Happy to read your thoughts, if the new behaviour doesn't work for you (and maybe others) or is too involved to implement, we can surely add a configuration option for that!

haasn:
[...]

PR #1129, opened 2022-10-17
docs: add pydbus playback status example

haasn:
I use this in my own setup becaue I find the ~1 second latency of the `playerctl` command annoying. This script bypasses that.

Note: Depends on #1128 to be fixed before it will work with current git master.


eladyn:
Thank you the PR! This page is currently not published / included in the book (as it is kind of a left-over from the wiki). If you'd like, you could clean up the user-supplied scripts and add that page into the book hierarchy, so that it is visible to users.

haasn:
Maybe we could just re-add it to the wiki instead?

haasn:
I'll close this for the time being since I don't care strongly about cleaning up the docs. I mostly just stumbled on that page via google, but google should hopefully also find this PR.

PR #1132, opened 2022-10-28
always emit the album art URL representing the largest image to DBus

l1na-forever:
This patch ensures we always return the clearest possible album art URL as part of the MPRIS event sent over DBus. I noticed this after [mpris-notifier](https://github.com/l1na-forever/mpris-notifier) started showing blurry album art. Upon further inspection, the first album art item in the list returned by Spotify had become the smallest image's URL, rather than the largest.

Before:
screenshot_20221028_005919254815550

After:
screenshot_20221028_005855845708237

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -776,7 +776,7 @@ fn insert_metadata(m: &mut HashMap<String, Variant<Box<dyn RefArg>>>, item: Play
         Variant(Box::new(
             item.images
                 .into_iter()
-                .next()
+                .max_by_key(|i| i.width.ok_or(0))
                 .map(|i| i.url)
                 .unwrap_or_default(),
         )),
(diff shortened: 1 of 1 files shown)


src/dbus_mpris.rs
```diff
@@ -776,7 +776,7 @@ fn insert_metadata(m: &mut HashMap<String, Variant<Box<dyn RefArg>>>, item: Play
         Variant(Box::new(
             item.images
                 .into_iter()
-                .next()
+                .max_by_key(|i| i.width.ok_or(0))
```
eladyn, line 779: Shouldn't that be:

```suggestion
                .max_by_key(|i| i.width.unwrap_or(0))
```

eladyn (changes requested):
Thank you very much for this. That is how it always should have been!

Only one question:

src/dbus_mpris.rs
l1na-forever: Yup, it sound be :+1:

eladyn (approved):
Tested and working. Thank you!

Plecra (approved):
As it should be 😊 

Side note - It's a bit frustrating we can't test this, any ideas on how we could avoid regressions here (maybe spotify starts returning images with different levels of compression)?

Issue #1131, opened 2022-10-24
Document necessity of feature flags for install commands


msandy:
The wiki instructions indicate that to build with a particular set of features, users should run e.g. `cargo build --release --features pulseaudio_backend`. Later, the wiki says that installation should be done with `cargo install --path . --locked`. It is not clear that the latter will re-compile the package without the previously-specified feature flags. This may be obvious to seasoned rust developers, but as someone who's just trying to follow the instructions verbatim, this omission caused me a major headache until I figured out what the problem was.

Specifically, I was running `cargo build --release --features pulseaudio_backend` and then subsequently `cargo install --path . --locked`. This led to the following error on launch:

```
The application panicked (crashed).
Message:  Unknown backend: pulseaudio.
```

To fix it, I skipped calling `cargo build` at all, and just ran `cargo install --path . --locked --features pulseaudio_backend`

eladyn:
Thank you for that clear report! That should indeed be changed.
Would you like to contribute a patch to the documentation yourself?

eladyn:
Closing similarly to the associated PR.

PR #1137, opened 2022-11-10
[docs] add note about cargo feature flags

msandy:
Fix for issue #1131

### DIFF
--- a/docs/src/installation/README.md
+++ b/docs/src/installation/README.md
@@ -78,6 +78,8 @@ flags that are set by default in `Cargo.toml`.
 
 `Spotifyd` is split into a base package plus additional features that can be toggled on or off during compilation. Those can be split into two groups: The audio backend features that are responsible for playing back the music and additional functionality features, which enhance your experience using `spotifyd`.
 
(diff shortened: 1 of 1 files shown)


eladyn:
Thank you! I just realized that maybe the real problem is rather that the documentation is misleading by saying that

> To install the resulting binary, run
> 
> ```bash
> cargo install --path . --locked
> ```

because `cargo install` doesn't actually install the **built** binary, but rather compiles it first and then installs it to the correct location. Maybe we could change that to better reflect the actual situation? Would that have solved your problem as well?

Plecra (approved):
Documenting this as explicitly diverging from `make` conventions sounds good to me :) It'd be great to make further changes along the lines you're describing @eladyn

slondr:
Seconding eladyn's point here — the docs are straight up wrong with the "resulting" part, and would probably be better served by making the `cargo install` section more prominent/higher in the section as to not imply it's part of a different installation process. That should help solve the problem of folks arriving in the repo just trying to install the software as fast as possible.

As an aside, I looked up the current 0.3.4 rollout on repology, and I want to give a special shout-out to Arch, Funtoo, and Void for being the only three distro [families] that are already shipping the new version to users :D

Issue #1144, opened 2022-12-17
Compilation errors in src/dbus_mpris.rs


gleb-chipiga:
**Description**
Compilation errors in `src/dbus_mpris.rs`

**To Reproduce**
`cargo install spotifyd -F pulseaudio_backend,dbus_keyring,dbus_mpris`
or
`cargo install --git https://github.com/Spotifyd/spotifyd -F pulseaudio_backend,dbus_keyring,dbus_mpris`

**Logs**
<details><summary>Click to show logs</summary>

```
   Compiling spotifyd v0.3.4 (/home/gleb/.cargo/git/checkouts/spotifyd-e8542734951960f8/db25d12)
error[E0407]: method `_type_static` is not a member of trait `Id`
   --> src/dbus_mpris.rs:361:17
    |
361 | /                 fn _type_static() -> Type
362 | |                 where
363 | |                     Self: Sized,
364 | |                 {
365 | |                     unreachable!("never called");
366 | |                 }
    | |_________________^ not a member of trait `Id`

error[E0407]: method `from_id_unchecked` is not a member of trait `Id`
   --> src/dbus_mpris.rs:368:17
    |
368 | /                 unsafe fn from_id_unchecked(_id: &str) -> Self
369 | |                 where
370 | |                     Self: Sized,
371 | |                 {
372 | |                     unreachable!("never called");
373 | |                 }
    | |_________________^ not a member of trait `Id`

error[E0404]: expected trait, found enum `PlayContextId`
   --> src/dbus_mpris.rs:350:41
    |
350 |             struct AnyContextId(Box<dyn PlayContextId>);
    |                                         ^^^^^^^^^^^^^ not a trait

error[E0404]: expected trait, found enum `PlayContextId`
   --> src/dbus_mpris.rs:375:18
    |
375 |             impl PlayContextId for AnyContextId {}
    |                  ^^^^^^^^^^^^^ not a trait

error[E0404]: expected trait, found enum `PlayableId`
   --> src/dbus_mpris.rs:378:34
    |
378 |                 Playable(Box<dyn PlayableId>),
    |                                  ^^^^^^^^^^ not a trait

Some errors have detailed explanations: E0404, E0407.
For more information about an error, try `rustc --explain E0404`.
error: could not compile `spotifyd` due to 5 previous errors
[...]

PR #1079, opened 2022-04-16
Upgrade rspotify to 0.11.5

eladyn:
This mainly upgrades the `rspotify` crate to `0.11.5`. Since large parts of the API surface changed between (the previous) `0.8.0` and `0.11.5`, the changes are a bit more complex in some cases.

## Other changes

- no longer clones the token many times on the creation of the dbus server and instead passes an `Arc` to the API client; this facilitates requesting a new token after expiry (fixes #1074)
- avoids cloning in the `.insert_metadata()` as much as possible
- makes the `rspotify` dependency optional (only needed with `dbus_mpris`)
- fix `device_name` matching in some dbus methods, as the `device.name` is not (or no longer?) percent encoded
- `cargo update`, since otherwise `spotifyd` would not compile with the dependency updates

## TODO

- [x] ~~improve clunky `OpenURI` implementation (see https://github.com/ramsayleung/rspotify/issues/218#issuecomment-1097240398)~~
   There is currently not much, we can do about that ugly implementation. Will probably be resolved in a future version of `rspotify`.
- [X] handle `Episode` cases properly
- [X] ~~consider using the `async` variant of the `rspotify` crate (`reqwest` client)~~ This is not easily feasible as it would require many changes that would make this PR quite difficult to review, better suited for a follow-up.
- [x] clean up the code
- [x] strip down requested token scopes

fixes #1069

PR #1145, opened 2022-12-17
remove workaround for rspotify id types

eladyn:
This upgrades `rspotify` to `0.11.6` and removes the workaround that was introduced in #1079, for which there is now much better library support.

related: #1144

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust: [stable, 1.62.0]
+        rust: [stable, 1.64]
         os: [macos-latest, ubuntu-latest]
         include:
           - os: macos-latest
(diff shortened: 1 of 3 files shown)


slondr (dismissed):
cool!

eladyn:
I rebased the PR on current master. Maybe we can get this merged soon-ish, such that CI passes again?

Icelk:
@eladyn maybe we should add --locked in CI? That fixes the CI, since it uses the versions from Cargo.lock

eladyn:
@Icelk That might indeed be a good idea. Well, the effort of fixing CI failures is – in theory – not that big. It's only becoming a hurdle, when other CI for other PRs are failing due to the fix not yet being merged.

slondr (approved):
🚀

eladyn:
Ugh, somehow it now suddenly wants to use `time@0.3.20`, which requires an MSRV bump. Should I do that here or should we just merge this and try to get the MSRV bump in a different PR as soon as possible?

Icelk:
I'll finish up #1182 today, which includes the MSRV bump. It would block on this, so if we merge the two right after each other, everything should work out.

slondr:
aight let's send it yeet

slondr (approved):

PR #1158, opened 2023-01-07
Set conditional default backend

fxsalazar:


eladyn:
Thank you! This is something I wanted to do for quite some time now.

While this solution is already much better than what has been there before, maybe we could go for an even "smarter" default backend selection logic. I'm thinking of something like this:

```rust
const fn default_backend() -> Backend {
  #[cfg(all(feature = "portaudio_backend", target_os="macos")]
  return Backend::Portaudio;
  return Bacend::Alsa;
  #[cfg(feature = "portaudio_backend")]
  return Backend::Portaudio;

  todo!("other backends");
  
  #[cfg(not(any(feature = "pulseaudio_backend", feature = "portaudio_backend", ...)))]
  compile_error!("Please enable at least one backend!");
}
```

This would prevent things like this:

```
The application panicked (crashed).
Message:  Unknown backend: alsa.
Location: src/setup.rs:169
```

Although this code is only a rough sketch and quite verbose / ugly.

If you don't want to look into this further, that would be fine as well, and we can just merge this version! :)

fxsalazar:
Yes, sure, I'll take a look to this approach.

fxsalazar:
I decided to with the `BACKEND_VALUES` which is already aggregating the features values. 🤔  does it make sense?

src/config.rs
```diff
@@ -39,11 +39,9 @@ pub enum Backend {
     Rodio,
 }
 
-const DEFAULT_BACKEND: Backend = if cfg!(target_os = "macos") {
-    Backend::PortAudio
-} else {
-    Backend::Alsa
-};
+fn default_backend() -> Backend {
+    return Backend::from_str(BACKEND_VALUES.first().unwrap()).unwrap_or(Backend::Alsa);
```
eladyn, line 43: To prevent the first `.unwrap()` from panicking at runtime, maybe we could add something like

```rust
#[cfg(not(any(feature = "pulseaudio_backend", feature = "portaudio_backend", feature = "alsa_backend", feature = "rodio_backend")))]
compile_error!("At least one of the backend features is required!");
```

The second `.unwrap_or(Backend::Alsa)` can IMO be converted to a simple `.unwrap`, since this should never fail.

eladyn (commented):
Oh, right. That seems like a good idea! Apart from a small comment, LGTM!

eladyn (approved):
Thank you!

slondr (approved):

PR #1162, opened 2023-01-14
New DBUS method: TransferPlayback

dzubybb:
This is a new TransferPlayback method available for DBUS. It's useful in my setup where spotifyd is part of audio system based on dietpi micro computer. It's used to transfer playback from other active spotify device.

Also i discovered weird error when setting my spotifyd device type to S_T_B. It was returned from device() method of spotify client as STB (all capitals) and was throwing error:
 ```
 json parse error: unknown variant `STB`, expected one of `Computer`, `Tablet`, `Smartphone`, `Speaker`, `Tv`, `Avr`, `Stb`, `AudioDongle`, `GameConsole`, `CastVideo`, `CastAudio`, `Automobile`, `Unknown` at line 16 column 18
```
 I didn't find `STB` anywhere in code so i assume that spotify changes that somehow ?

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -21,7 +21,7 @@ use librespot_core::{
     spotify_id::SpotifyAudioType,
 };
 use librespot_playback::player::PlayerEvent;
-use log::{error, info};
+use log::{error, info, warn};
 use rspotify::{
     model::{
         offset::Offset, AlbumId, ArtistId, EpisodeId, IdError, PlayableItem, PlaylistId,
@@ -413,15 +413,7 @@ async fn create_dbus_server(
 
             let uri = Uri::from_id(id_type, id).map_err(|_| MethodErr::invalid_arg(&uri))?;
 
-            let device_id = sp_client.device().ok().and_then(|devices| {
-                devices.into_iter().find_map(|d| {
-                    if d.is_active && d.name == mv_device_name {
-                        Some(d.id)
-                    } else {
-                        None
-                    }
-                })
-            });
+            let device_id = get_device_id(&sp_client, &mv_device_name, true);
 
             if let Some(device_id) = device_id {
                 match uri {
@@ -444,6 +436,24 @@ async fn create_dbus_server(
                 }
             }
             Ok(())
+        });
+
+        let mv_device_name = device_name.clone();
+        let sp_client = Arc::clone(&spotify_api_client);
+        b.method("TransferPlayback", (), ("status", ), move |_, _, (): ()| {
+            let device_id = get_device_id(&sp_client, &mv_device_name, false).flatten();
+            let status = if let Some(device_id) = device_id {
+                info!("Transferring playback to device {}", device_id);
+                match sp_client.transfer_playback(&device_id, Some(true)) {
+                    Ok(_) => "Playback transferred".to_string(),
+                    Err(err) => format!("Error {}", err)
+                }
+            } else {
+                let msg = format!("Could not find device with name {}", mv_device_name);
+                warn!("{}", msg);
+                msg
+            };
+            Ok((status, ))
         });
 
         let mv_device_name = device_name.clone();
@@ -696,6 +706,24 @@ async fn create_dbus_server(
             conn.send(msg).unwrap();
         }
     }
+}
+
+fn get_device_id(sp_client: &Arc<AuthCodeSpotify>, device_name: &String, only_active: bool) -> Option<Option<String>> {
+    let device_result = sp_client.device();
+    return match device_result {
+        Ok(devices) => devices.into_iter().find_map(|d| {
+            if d.name.eq(device_name) && (d.is_active || !only_active)  {
+                info!("Found device: {}, active: {}", d.name, d.is_active);
+                Some(d.id)
+            } else {
+                None
+            }
+        }),
+        Err(err) => {
+            error!("Get devices error: {}", err);
+            None
+        }
+    };
 }
 
 fn uri_to_object_path(uri: String) -> dbus::Path<'static> {
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -457,7 +457,9 @@ async fn create_dbus_server(
                 Ok("Stopped".to_string())
             });
 
+        let mv_device_name = device_name.clone();
         let sp_client = Arc::clone(&spotify_api_client);
+        let sp_client2 = Arc::clone(&spotify_api_client);
         b.property("Shuffle")
             .emits_changed_false()
             .get(move |_, _| {
@@ -467,6 +469,23 @@ async fn create_dbus_server(
                     .flatten()
                     .map_or(false, |p| p.shuffle_state);
                 Ok(shuffle_status)
+            })
+            .set(move |_, _, value| {
+                let device_id = get_device_id(&sp_client2, &mv_device_name, true);
+                if let Some(device_id) = device_id {
+                    match sp_client2.shuffle(value, Some(&device_id)) {
+                        Ok(_) => Ok(None),
+                        Err(err) => {
+                            let e = format!("SetShuffle failed: {}", err);
+                            error!("{}", e);
+                            Err(MethodErr::failed(&e))
+                        }
+                    }
+                } else {
+                    let msg = format!("Could not find device with name {}", mv_device_name);
+                    warn!("SetShuffle: {}", msg);
+                    Err(MethodErr::failed(&msg))
+                }
             });
 
         b.property("Rate").emits_changed_const().get(|_, _| Ok(1.0));
@@ -735,7 +754,7 @@ async fn create_dbus_server(
 
 fn get_device_id(
     sp_client: &Arc<AuthCodeSpotify>,
-    device_name: &String,
+    device_name: &str,
     only_active: bool,
 ) -> Option<String> {
     let device_result = sp_client.device();
@@ -749,6 +768,17 @@ fn get_device_id(
             }
         }),
         Err(err) => {
+            let expires_at = sp_client
+                .get_token()
+                .lock()
+                .ok()
+                .unwrap()
+                .as_ref()
+                .unwrap()
+                .expires_at
+                .unwrap();
+            info!("Token expires at: {}", expires_at);
+
             error!("Get devices error: {}", err);
             None
         }
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -568,20 +568,23 @@ async fn create_dbus_server(
 
         let mv_device_name = device_name.clone();
         let sp_client = Arc::clone(&spotify_api_client);
-        b.method("TransferPlayback", (), ("status", ), move |_, _, (): ()| {
+        b.method("TransferPlayback", (), (), move |_, _, (): ()| {
             let device_id = get_device_id(&sp_client, &mv_device_name, false).flatten();
-            let status = if let Some(device_id) = device_id {
+            if let Some(device_id) = device_id {
                 info!("Transferring playback to device {}", device_id);
                 match sp_client.transfer_playback(&device_id, Some(true)) {
-                    Ok(_) => "Playback transferred".to_string(),
-                    Err(err) => format!("Error {}", err)
+                    Ok(_) =>Ok(()),
+                    Err(err) => {
+                        let e = format!("RSpotify error: {}", err);
+                        error!("{}", e);
+                        Err(MethodErr::failed(&e))
+                    },
                 }
             } else {
                 let msg = format!("Could not find device with name {}", mv_device_name);
                 warn!("{}", msg);
-                msg
-            };
-            Ok((status, ))
+                Err(MethodErr::failed(&msg))
+            }
         });
     });
 
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -234,12 +234,14 @@ async fn create_dbus_server(
         b.method("VolumeUp", (), (), move |_, _, (): ()| {
             local_spirc.volume_up();
             Ok(())
-        }).deprecated();
+        })
+        .deprecated();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by eladyn** [comment]
Thank you for the contribution!

In general, looks like a good idea. I only just had a quick look and didn't look at the code properly yet, but one concern that I have, is that the interface we'd be adding that method to is `org.mpris.MediaPlayer2.Player`, which has a formal specification which doesn't include that method. We already differ from that spec with the `VolumeUp` and `VolumeDown` methods, but IMO we should try to stick to that as much as possible. Maybe you could add a separate interface specific to `spotifyd` to the dbus server which would then include your new method and maybe the two others that also don't fit into the spec.
Open to your opinion on this, though!

---

Regarding the error you're receiving, I'm quite sure that this comes from `raspotify` failing to parse the Web API answer. The reason is probably a missing alias [here](https://github.com/ramsayleung/rspotify/blob/f038a2db1182f8a71bc5b067fe5f6be80c779cb4/rspotify-model/src/enums/types.rs#L91-L113). If you want you could either contribute a fix to them directly or report it over there so that they can fix it. :)

**Comment by dzubybb** [comment]
> Maybe you could add a separate interface specific to spotifyd to the dbus server which would then include your new method and maybe the two others that also don't fit into the spec.

Yes, i think this is good idea. Currently there are two DBUS interfaces `org.mpris.MediaPlayer2` and `org.mpris.MediaPlayer2.Player`, both on `/org/mpris/MediaPlayer2` path. 
Maybe something like interface `org.spotifyd.Controls` on path `/org/spotifyd/Controls` ? Or do You have better idea ?

>Regarding the error you're receiving, I'm quite sure that this comes from raspotify failing to parse the Web API answer. The reason is probably a missing alias [here](https://github.com/ramsayleung/rspotify/blob/f038a2db1182f8a71bc5b067fe5f6be80c779cb4/rspotify-model/src/enums/types.rs#L91-L113).

I will try to fix it.

Also i found another error, when trying to call my new method after longer time (about 20 hours, service was running this whole time) i got `Get devices error: http error: status code 401` on call `sp_client.device()`, looks like token had expired. Shouldn't it be refreshed automatically ?

**Comment by dzubybb** [comment]
>If you want you could either contribute a fix to them directly or report it over there so that they can fix it. :)

Issue https://github.com/ramsayleung/rspotify/issues/383 pull request https://github.com/ramsayleung/rspotify/pull/384

**Comment by eladyn** [comment]
> Yes, i think this is good idea. Currently there are two DBUS interfaces `org.mpris.MediaPlayer2` and `org.mpris.MediaPlayer2.Player`, both on `/org/mpris/MediaPlayer2` path. Maybe something like interface `org.spotifyd.Controls` on path `/org/spotifyd/Controls` ? Or do You have better idea ?

Sounds good. To stick to a domain that the project actually "owns", maybe we could use `io.github.spotifyd.Controls`?

> Also i found another error, when trying to call my new method after longer time (about 20 hours, service was running this whole time) i got `Get devices error: http error: status code 401` on call `sp_client.device()`, looks like token had expired. Shouldn't it be refreshed automatically ?

Oh, that's weird! I think I implemented auto-refresh for the tokens some time ago and the D-Bus things should only be called, when a non-expired token is available? No idea, what's happening there. Did you notice anything special logged in the output?

**Comment by dzubybb** [comment]
> Sounds good. To stick to a domain that the project actually "owns", maybe we could use io.github.spotifyd.Controls?

I've pushed changes, for now i copied VolumeUp and VolumeDown methods, instead of moving them, to avoid breaking changes. I've also marked those methods on `org.mpris.MediaPlayer2.Player` interface as deprecated. 

> Oh, that's weird! I think I implemented auto-refresh for the tokens some time ago and the D-Bus things should only be called, when a non-expired token is available? No idea, what's happening there. Did you notice anything special logged in the output?

I didn't notice anything else in logs, but i added logging of token expiration time, so i will know more when error repeats itself.

File: src/dbus_mpris.rs [hunk 1]
```diff
@@ -562,12 +554,49 @@ async fn create_dbus_server(
...
+                match sp_client.transfer_playback(&device_id, Some(true)) {
+                    Ok(_) => "Playback transferred".to_string(),
+                    Err(err) => format!("Error {}", err)
+                }
+            } else {
+                let msg = format!("Could not find device with name {}", mv_device_name);
+                warn!("{}", msg);
+                msg
+            };
+            Ok((status, ))
```
eladyn, line 584: Instead of returning a string here, what do you think about reusing the built-in dbus error variants?
This would mean:
- return nothing (aka `Ok(())`) in case of success.
- return an error message wrapped in [`Err(dbus::MethodErr)`](https://docs.rs/dbus/latest/dbus/struct.MethodErr.html#method.failed) in case of failure

I guess that might be more dbus-like? My experience with interface design over there is limited, however, so I might be wrong.

src/dbus_mpris.rs
```diff
@@ -698,6 +727,24 @@ async fn create_dbus_server(
     }
 }
 
+fn get_device_id(sp_client: &Arc<AuthCodeSpotify>, device_name: &String, only_active: bool) -> Option<Option<String>> {
+    let device_result = sp_client.device();
+    return match device_result {
```
eladyn, line 732: You can use `match` directly and omit the semicolon at the end, and it should work as well. ;)

src/dbus_mpris.rs
```diff
@@ -698,6 +727,24 @@ async fn create_dbus_server(
...
+        Ok(devices) => devices.into_iter().find_map(|d| {
+            if d.name.eq(device_name) && (d.is_active || !only_active)  {
+                info!("Found device: {}, active: {}", d.name, d.is_active);
+                Some(d.id)
```
eladyn, line 736: I haven't dug into the code. Is there a reason, why we're returning a doubly wrapped `Option` instead of just `flatten`ing it?

eladyn (commented):
Thank you. Some comments, but apart from that: looks good!

slondr:
> To stick to a domain that the project actually "owns", maybe we could use `io.github.spotifyd.Controls`?

Do we want to grab an actual domain name for the project? Seems useful in general, and for this specifically. I'd be happy to set that up if so.

src/dbus_mpris.rs
dzubybb: I don't see any, fixed.

src/dbus_mpris.rs
dzubybb: Fixed

src/dbus_mpris.rs
dzubybb: Done.

dzubybb:
@eladyn It looks like token is not always refreshed. I added code to log token expiration time an got those logs :
```
Jan 26 15:38:24 DietPi spotifyd[774]: Token expires at: 2023-01-26 14:20:53.258021114 UTC
Jan 26 15:38:24 DietPi spotifyd[774]: Get devices error: http error: status code 401
Jan 26 15:38:24 DietPi spotifyd[774]: Could not find device with name unitra_dietpi
Jan 26 15:38:38 DietPi spotifyd[774]: Token expires at: 2023-01-26 14:20:53.258021114 UTC
Jan 26 15:38:38 DietPi spotifyd[774]: Get devices error: http error: status code 401
Jan 26 15:38:38 DietPi spotifyd[774]: Could not find device with name unitra_dietpi
Jan 26 15:39:32 DietPi spotifyd[774]: Token expires at: 2023-01-26 14:20:53.258021114 UTC
Jan 26 15:39:32 DietPi spotifyd[774]: Get devices error: http error: status code 401
Jan 26 15:39:32 DietPi spotifyd[774]: Could not find device with name unitra_dietpi
```
Log time is UTC +1, so it's about 18 minutes past token expiration.

eladyn:
@dzubybb Oh, right. This took me a bit, to figure out! What's happening here, is that we spawn our dbus handling logic on a separate tokio task. This way, our main logic isn't blocked by those sync http requests (maybe there are other reasons as well, I don't know), but this also means that our token refresh logic is not run, when there's activity in the dbus server, but only, when there are new events for our PropertiesChanged signal. That is, why it was always working for me and for you it regularly failed: I only tested the behaviour with `spotifyd` continuously running while you had longer periods without any playback which caused the token to expire. And since you aren't able to re-enable playback, no activity occurs and the token is still not refreshed.

I'm not yet sure, what an elegant solution to this might look like, and I'm likely not going to get to it in the next few days, but now we have at least identified the root cause.

Apart from that, I hope that get around to taking a look at your updated code in the following week.

eladyn:
@slondr Sorry for not replying earlier! In general, I like the idea of having a proper domain for `spotifyd`. I'm not sure, however, if the benefits weigh up against the increased maintenance and cost burden on the domain holders (e.g. you).

My concern in that regard would be, that as soon as such a domain is established, dependencies on that domain will be popping up (e.g. usage of it as DBus name, linking to it in README, issues and other resources). Some of these are reverted more easily than others, so such a decision shouldn't be made light-heartedly, and the ones taking responsibility for the domain should be aware of the possible implications. But if you'd like to take up that challenge, I won't be stopping you. :)

src/dbus_mpris.rs
```diff
@@ -562,12 +554,52 @@ async fn create_dbus_server(
...
+                    Ok(_) =>Ok(()),
+                    Err(err) => {
+                        let e = format!("RSpotify error: {}", err);
+                        error!("{}", e);
+                        Err(MethodErr::failed(&e))
+                    },
+                }
+            } else {
+                let msg = format!("Could not find device with name {}", mv_device_name);
+                warn!("{}", msg);
```
eladyn, line 585:
```suggestion
                warn!("TransferPlayback: {}", msg);
```

Maybe, we could add some context here?

src/dbus_mpris.rs
```diff
@@ -562,12 +554,52 @@ async fn create_dbus_server(
...
+        let mv_device_name = device_name.clone();
+        let sp_client = Arc::clone(&spotify_api_client);
+        b.method("TransferPlayback", (), (), move |_, _, (): ()| {
+            let device_id = get_device_id(&sp_client, &mv_device_name, false);
+            if let Some(device_id) = device_id {
+                info!("Transferring playback to device {}", device_id);
+                match sp_client.transfer_playback(&device_id, Some(true)) {
+                    Ok(_) =>Ok(()),
+                    Err(err) => {
+                        let e = format!("RSpotify error: {}", err);
```
eladyn, line 578:
```suggestion
                        let e = format!("TransferPlayback failed: {}", err);
```

I'd personally avoid naming the used crate here, as it doesn't give the user much information.

src/dbus_mpris.rs
eladyn: Oh, my bad. Instead of flattening it, it should probably just work by removing this `Some`.

src/dbus_mpris.rs
```diff
@@ -698,6 +730,24 @@ async fn create_dbus_server(
     }
 }
 
+fn get_device_id(sp_client: &Arc<AuthCodeSpotify>, device_name: &String, only_active: bool) -> Option<String> {
```
eladyn, line 733:
```suggestion
fn get_device_id(sp_client: &Arc<AuthCodeSpotify>, device_name: &str, only_active: bool) -> Option<String> {
```

src/dbus_mpris.rs
```diff
@@ -21,7 +21,7 @@ use librespot_core::{
     spotify_id::SpotifyAudioType,
```
eladyn, line 21: Before we can merge this, could you run `cargo fmt` and `cargo clippy`, so that CI is happy? :)

eladyn (changes requested):
Some minor comments, other than that, this looks good to me. And it's working nicely as well!

src/dbus_mpris.rs
eladyn: Oh, my bad. Instead of flattening it, it should probably just work by removing this `Some`.

slondr:
> @slondr Sorry for not replying earlier! In general, I like the idea of having a proper domain for `spotifyd`. I'm not sure, however, if the benefits weigh up against the increased maintenance and cost burden on the domain holders (e.g. you).
> 
> My concern in that regard would be, that as soon as such a domain is established, dependencies on that domain will be popping up (e.g. usage of it as DBus name, linking to it in README, issues and other resources). Some of these are reverted more easily than others, so such a decision shouldn't be made light-heartedly, and the ones taking responsibility for the domain should be aware of the possible implications. But if you'd like to take up that challenge, I won't be stopping you. :)

Oh it's no big deal for me, I have close to a dozen domains already that I don't even do anything with (but definitely will one day! I swear!), so adding 1 to the pile is fine.

Looks like I can grab spotifyd.rs for like $25 a year!

dzubybb:
Code fixes done. I also added setter for Shuffle property. Should i commit it here, or open new pull request for that ?

eladyn:
> Code fixes done.

Thanks, it only seems that you missed one item? Or was this intentional / doesn't work like that?

> I also added setter for Shuffle property. Should i commit it here, or open new pull request for that ?

Oh, that's great! I think, opening a new pull request should make it easier to review / allow us to merge this one earlier. It would be totally fine by me, as well, if you just based that second PR on the changes in this one. This way, you don't have to go through the hassle and rebase them on master.

dzubybb:
Damn, pushed too much :)

dzubybb:
> It would be totally fine by me, as well, if you just based that second PR on the changes in this one.

So... how do i do that ? Just create another PR from same master branch in my fork ?

eladyn:
> So... how do i do that ? Just create another PR from same master branch in my fork ?

You can create a new branch from the current commit on your master branch, add your changes and push that branch to your forked repo. Then, you should be able to create a new pull request from this second branch into this repo's master.

eladyn (commented):
Two little last things that I missed previously. After that, I think we're good to go!

eladyn (approved):
Thank you! LGTM and works fine.

slondr (approved):
epic

PR #1172, opened 2023-03-02
update badges and add link to matrix

eladyn:
This removes/updates some obsolete badges and adds a new one for the matrix room. ([Preview](https://github.com/Spotifyd/spotifyd/tree/badge-updates#readme))

Justification for the individual updates / removals:
- contributors are also listed on the right side of the repo's front page and haven't been updated in a while
- cargo downloads don't accurately represent the actual downloads of this project, which probably happen more via GitHub releases or other distributions (the shown number is less than the amount of stars this repo has)
- the CI and CD badge didn't match the style of the others, so I went for the ones from shields.io

Now, when we're already at it, would you prefer a [different style](https://shields.io/#styles) for the badges?

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: General questions and support
+    url: https://matrix.to/#/#spotifyd:matrix.org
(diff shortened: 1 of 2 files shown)


slondr (approved):
LGTM!

With regards to styles: I think they all look fine, though if we do want to change it, flat-square looks the nicest to me.

eladyn:
Thanks!

PR #1174, opened 2023-03-02
update keyring crate

eladyn:
This updates to the newest release of the keyring crate and updates the messages of the keyring query. From my testing, this is backwards compatible with the previous version – at least the parts that we were using.

Fixes #1086. (at least the last update there was, that changes in keyring, which have landed, might resolve the issue)

PR #1179, opened 2023-03-03
Create dependabot.yml

slondr:
This enables the Dependabot feature which submits PRs for dep upgrades even when they aren't security-related.

I have no idea if this would be useful or not but I figure hey why not try it out

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: "monthly"


eladyn:
In general, I like having some workflow that regularly updates the dependencies. Otherwise, in PRs that need it (e.g., #1174) I have to run `cargo update` and end up having massive line change counts.

However, I'm not very happy with the way that dependabot solves this (individual PR's for each dependency) as this creates a lot of noise in PRs and especially since we have this 2-reviewer policy isn't as easy to act on as it should be. I'd rather have it like it was proposed [here](https://github.com/dependabot/dependabot-core/issues/2174), since this project depends on a massive amount of crates and weekly batch updates would be much more manageable.

Apart from that, I don't think updating `Cargo.lock` does have major advantages, since `cargo` seems to somehow ignore the file anyway except when using `--locked`. (at least from my experience)

But if you have some advantages in mind that I didn't think of, please let me know. :)

slondr:
Hey, fair enough. I sub'd to that dependabot issue - we can continue to handle this manually for now.

PR #1182, opened 2023-03-08
Librespot upgrade to 0.4

Icelk:
Upgrades librespot to 0.4

# Considerations

- Audio normalization: I'm unsure of the changes, but I think we could improve clipping with normilazation when using softvol
- Credentials: librespot now seems to want to handle credentials storage by itself. Should we hand over this job?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,11 @@ tokio = {version = "1.6.1", features = ["signal", "rt-multi-thread", "process", 
 tokio-stream = "0.1.7"
 url = "2.2.2"
 xdg = "2.2"
-librespot-audio = { version = "0.2.0", default-features=false, features = ["with-tremor"] }
-librespot-playback = { version = "0.2.0", default-features=false }
-librespot-core = { version = "0.2.0"}
-librespot-connect = { version = "0.2.0"}
+librespot-audio = { version = "0.4", default-features = false }
+librespot-playback = { version = "0.4", default-features = false }
+librespot-core = { version = "0.4" }
+librespot-discovery = { version = "0.4" }
+librespot-connect = { version = "0.4" }
 toml = "0.7"
 color-eyre = "0.6"
 
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -614,12 +614,12 @@ async fn create_dbus_server(
                 if let Some(track_id) = track_id {
                     let item = match track_id.audio_type {
                         SpotifyAudioType::Track => {
-                            let track_id = TrackId::from_id(track_id.to_base62()).unwrap();
+                            let track_id = TrackId::from_id(track_id.to_base62().unwrap()).unwrap();
                             let track = spotify_api_client.track(track_id).map(PlayableItem::Track);
                             Some(track)
                         }
                         SpotifyAudioType::Podcast => {
-                            let id = EpisodeId::from_id(track_id.to_base62()).unwrap();
+                            let id = EpisodeId::from_id(track_id.to_base62().unwrap()).unwrap();
                             let episode = spotify_api_client
                                 .get_an_episode(id, None)
                                 .map(PlayableItem::Episode);
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -8,13 +8,14 @@ use futures::{
     stream::Peekable,
     Future, FutureExt, StreamExt,
 };
-use librespot_connect::{discovery::DiscoveryStream, spirc::Spirc};
+use librespot_connect::spirc::Spirc;
 use librespot_core::{
     authentication::Credentials,
     cache::Cache,
-    config::{ConnectConfig, DeviceType, SessionConfig, VolumeCtrl},
+    config::{ConnectConfig, DeviceType, SessionConfig},
     session::{Session, SessionError},
 };
+use librespot_discovery::Discovery;
 use librespot_playback::{
     audio_backend::Sink,
     config::{AudioFormat, PlayerConfig},
@@ -38,20 +39,20 @@ pub struct SpotifydState {
 }
 
 pub(crate) enum CredentialsProvider {
-    DiscoveryStream(Peekable<DiscoveryStream>),
+    Discovery(Peekable<Discovery>),
     SpotifyCredentials(Credentials),
 }
 
-impl From<DiscoveryStream> for CredentialsProvider {
-    fn from(stream: DiscoveryStream) -> Self {
-        CredentialsProvider::DiscoveryStream(stream.peekable())
+impl From<Discovery> for CredentialsProvider {
+    fn from(stream: Discovery) -> Self {
+        CredentialsProvider::Discovery(stream.peekable())
     }
 }
 
 impl CredentialsProvider {
     async fn get_credentials(&mut self) -> Credentials {
         match self {
-            CredentialsProvider::DiscoveryStream(stream) => stream.next().await.unwrap(),
+            CredentialsProvider::Discovery(stream) => stream.next().await.unwrap(),
             CredentialsProvider::SpotifyCredentials(creds) => creds.clone(),
         }
     }
@@ -59,7 +60,7 @@ impl CredentialsProvider {
     // wait for an incoming connection if the underlying provider is a discovery stream
     async fn incoming_connection(&mut self) {
         match self {
-            CredentialsProvider::DiscoveryStream(stream) => {
+            CredentialsProvider::Discovery(stream) => {
                 let peeked = Pin::new(stream).peek().await;
                 if peeked.is_none() {
                     future::pending().await
@@ -76,7 +77,7 @@ pub(crate) struct MainLoop {
     pub(crate) player_config: PlayerConfig,
     pub(crate) session_config: SessionConfig,
     pub(crate) autoplay: bool,
-    pub(crate) volume_ctrl: VolumeCtrl,
+    pub(crate) has_volume_ctrl: bool,
     pub(crate) initial_volume: Option<u16>,
     pub(crate) shell: String,
     pub(crate) device_type: DeviceType,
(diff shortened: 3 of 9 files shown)


Icelk:
This seems to be blocking on rust 1.63. Tomorrow, a new rust version will release, so maybe we could up our required version? (if you have a MSRV policy of 5 stable releases back)

eladyn:
Oh, cool! I only had a quick look and am surprised that the amount of changes is not as big as I feared (given the major version jump 0.2 &rarr; 0.4).

> This seems to be blocking on rust 1.63. Tomorrow, a new rust version will release, so maybe we could up our required version? (if you have a MSRV policy of 5 stable releases back)

I don't think, we have a strict MSRV policy. The closest to such a thing is probably https://github.com/Spotifyd/spotifyd/issues/1123#issuecomment-1272615757. I think, bumping to 1.63 or 1.64 should be totally fine. You'll just need to change that in `Cargo.toml` as well as in CI.

Icelk:
> Oh, cool! I only had a quick look and am surprised that the amount of changes is not as big as I feared (given the major version jump 0.2 → 0.4).

I know! Hopefully, once v0.5 lands, the upgrade will be easier.

> I don't think, we have a strict MSRV policy. The closest to such a thing is probably https://github.com/Spotifyd/spotifyd/issues/1123#issuecomment-1272615757. I think, bumping to 1.63 or 1.64 should be totally fine. You'll just need to change that in Cargo.toml as well as in CI.

Nice, I'll just do that. The 1.62 recommendation came from a long time ago, so I think it's OK to bump two versions (12 weeks).

slondr:
Just to weigh in on the MSRV discussion: my suggestion of using Gentoo Stable as a heuristic would currently put us at Rust 1.67; of course, if we don't actively have a reason to go beyond a specific version, keeping to an older version is fine. 

So, I'm ok with us bumping to 1.64 for this PR.

slondr (changes requested):
The changes made look good to me, and the checks are passing.

@Icelk, could you address the merge conflicts? They're from the "none" audio controller commit, I believe. 

W/r/t the normalization or credentials features, I'd be interested to see what those would look like in practice. For credentials specifically it's interesting to imagine some of our codebase downsizing, but naturally we'd want to ensure we aren't losing functionality.

Icelk:
> @Icelk, could you address the merge conflicts? They're from the "none" audio controller commit, I believe.

Yes, I just did :)

> W/r/t the normalization or credentials features, I'd be interested to see what those would look like in practice. For credentials specifically it's interesting to imagine some of our codebase downsizing, but naturally we'd want to ensure we aren't losing functionality.

Once this is merged, I can try removing our credentials logic and share the process. The cache format / persistence doesn't need to be defined? Because we get the password and username on startup, so the cache is only for speed? Why don't we just keep it in memory?

slondr (approved):

src/dbus_mpris.rs
```diff
@@ -349,48 +349,21 @@ async fn create_dbus_server(
...
-                {
-                    unreachable!("never called");
-                }
-            }
-            impl PlayContextId for AnyContextId {}
-
             enum Uri {
-                Playable(Box<dyn PlayableId>),
-                Context(AnyContextId),
+                Playable(PlayableId<'static>),
```
eladyn, line 353: Regarding these changes in `OpenUri`, there's #1145. That also simplifies this code, which can be cleaned up even more due to improvements in `rspotify`. :)

src/main_loop.rs
```diff
@@ -94,7 +96,9 @@ impl MainLoop {
         let session_config = self.session_config.clone();
         let cache = self.spotifyd_state.cache.clone();
 
-        Session::connect(session_config, creds, cache).await
+        Session::connect(session_config, creds, cache, false)
+            .await
+            .map(|(session, _creds)| session)
```
eladyn, line 101:
```suggestion
            .0
```

Should also work, or am I missing something?

src/setup.rs
```diff
@@ -21,11 +18,17 @@ use log::info;
 use std::str::FromStr;
 
 pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLoop {
-    let mut mixer = {
+    let mixer_config = MixerConfig {
+        device: config.device_name.clone(),
+        volume_ctrl: librespot_playback::config::VolumeCtrl::default(),
+        control: "".to_owned(),
+        ..Default::default()
+    };
```
eladyn, line 26: I'm a bit confused by this. As far as I can tell, this is only used by the `SoftVol` mixer below. The code for this mixer actually only cares about the `volume_ctrl` field. And since you're just setting that to default here, wouldn't it be easier to just do `MixerConfig::default()` below?

Cargo.toml
```diff
@@ -33,10 +33,11 @@ tokio-compat-02 = "0.2.0"
...
 xdg = "2.2"
-librespot-audio = { version = "0.2.0", default-features=false, features = ["with-tremor"] }
-librespot-playback = { version = "0.2.0", default-features=false }
-librespot-core = { version = "0.2.0"}
-librespot-connect = { version = "0.2.0"}
+librespot-audio = { version = "0.4.0", default-features = false }
+librespot-playback = { version = "0.4.0", default-features = false }
+librespot-core = { version = "0.4.0" }
+librespot-discovery = { version = "0.4.0" }
+librespot-connect = { version = "0.4.0" }
```
eladyn, line 40: The most recent version of `librespot` seems to be `0.4.2`. Should we just omit the last number? Or have there been changes between `0.4.0` and `0.4.2`, such that this doesn't yet work with `0.4.2`?

src/main_loop.rs
```diff
@@ -119,13 +123,12 @@ impl MainLoop {
...
 
             let mixer = (self.audio_setup.mixer)();
-            let audio_filter = mixer.get_audio_filter();
             let backend = self.audio_setup.backend;
             let audio_device = self.audio_setup.audio_device.clone();
             let (player, mut event_channel) = Player::new(
                 self.player_config.clone(),
                 session.clone(),
-                audio_filter,
+                Box::new(NoOpVolume),
```
eladyn, line 131: I think, this needs to be 
```suggestion
                mixer.get_soft_volume(),
```
for `SoftVol` to work. (Otherwise, the `Player` has no idea, what the multiplication factor for the audio samples.)

src/main_loop.rs
```diff
@@ -136,8 +139,8 @@ impl MainLoop {
                     autoplay: self.autoplay,
                     name: self.spotifyd_state.device_name.clone(),
                     device_type: self.device_type,
-                    volume: self.initial_volume.unwrap_or_else(|| mixer.volume()),
-                    volume_ctrl: self.volume_ctrl.clone(),
+                    initial_volume: Some(self.initial_volume.unwrap_or_else(|| mixer.volume())),
```
eladyn, line 142: Now, this has become an `Option`, we should be able to safely pass just `self.initial_volume`, since `Spirc` is going to call the mixer itself anyway.

src/no_mixer.rs
```diff
@@ -1,23 +1,15 @@
-use librespot_playback::mixer::{AudioFilter, Mixer, MixerConfig};
+use librespot_playback::mixer::{Mixer, MixerConfig};
 
 pub struct NoMixer {}
```
eladyn, line 3: Only a minor nitpick:

```suggestion
pub struct NoMixer;
```

Didn't think of that in the other PR.

eladyn (changes requested):
Thank you again for working on this! I added some comments, but overall this looks good.

eladyn:
I noticed that `librespot` apparently also added an `AlsaMixer`, which seems to handle a lot of ALSA weirdness and many edge cases. Maybe we could switch to that at some point, to reduce our code size. (Doesn't need to be this PR, though.)

src/main_loop.rs
Icelk: Yes, because we're mapping a result, so we can't just access the first field of the tuple directly. Using `?` then requires us to wrap the expression in `Ok`. I think this is the most readable.

Cargo.toml
Icelk: I just omitted the last number, in accordance to #1183

src/setup.rs
Icelk: Good point. Though I wanted to make sure the device name was right and to be extra explicit, but using the default looks nicer.

src/main_loop.rs
Icelk: 👍🏾

src/dbus_mpris.rs
Icelk: As discussed in #1145, we'll merge that first and then I'll rebase this to remove my changes :)

Icelk:
> I noticed that librespot apparently also added an AlsaMixer, which seems to handle a lot of ALSA weirdness and many edge cases. Maybe we could switch to that at some point, to reduce our code size. (Doesn't need to be this PR, though.)

Another thing I could look into in the future :)

Icelk:
Thanks for the feedback! Everything you mentioned is fixed, and I rebased.

src/dbus_mpris.rs
slondr: Hmm, if this is blocking on #1145, but #1145 would break builds until this is landed, we could be in an sticky situation for a period of time which is probably avoidable. 

Maybe we should have both this PR and #1145 target a non-master branch, merge both into that branch, and then squash the 2 commits and merge into master.

Icelk: Good point. Or we could just bump the MSRV in the other PR.

eladyn: > Good point. Or we could just bump the MSRV in the other PR.

Sounds like it would be the easiest. I'll do that.

slondr:
@Icelk Now that #1183 and #1145 are landed—the latter containing the MSRV bump—could you rebase this on master one more time? I think we should be good to go thereafter

slondr (changes requested):
Now that https://github.com/Spotifyd/spotifyd/pull/1183 and https://github.com/Spotifyd/spotifyd/pull/1145 are landed—the latter containing the MSRV bump—could you rebase this on master one more time @Icelk? I think we should be good to go thereafter

Icelk:
@slondr Just rebased. Thanks for merging #1183 & #1145!

slondr (approved):
Awesome!

eladyn (approved):
Thank you, merging! :rocket:

Issue #1028, opened 2021-12-13
Spotifyd snap very old version


pietryszak:
**Description**
Please update spotifyd in snapcraft to never version. The current  is 0.2.3. It's a very old version

eladyn:
@SimonPersson, [snapcraft.io](https://snapcraft.io/spotifyd) reports you as the publisher for the snap. Maybe this could be done together with the upcoming release?

SimonPersson:
I could take a look at it, but my linux distro doesn't support Snap so it would be a hassle to test. Back when I published this, someone working for Canonical set up the build script and helped me test that it was working fine on Ubuntu.

Perhaps the best thing to do if we don't have anyone to maintain this is to just take down the snap? As it stands, that outdated snap gives a pretty bad impression of spotifyd.

eladyn:
Ah, I see.

> Perhaps the best thing to do if we don't have anyone to maintain this is to just take down the snap? As it stands, that outdated snap gives a pretty bad impression of spotifyd.

That seems like a good idea, since without anyone maintaining it, it doesn't make much sense having it as a supported installation method. As far as I can understand, it should still be possible for others to publish their own `spotifyd` snaps without "official" support by the project.

SimonPersson:
Done. [The link](https://snapcraft.io/spotifyd) still seems to work, but perhaps that's a caching issue on shapcraft's side. In any case, I can't find it via search anymore, nor is it listed under my snaps.

PR #1186, opened 2023-03-10
remove snap files and mentions

eladyn:
Since the [official snap has been taken down](https://github.com/Spotifyd/spotifyd/issues/1028#issuecomment-1248526421) some time ago, this removes the remaining leftovers in the repository.


slondr (approved):
Love it

PR #1187, opened 2023-03-11
add CNAME to gh-pages

eladyn:
This prevents the action to overwrite our custom domain name. See [here](https://github.com/marketplace/actions/github-pages-action#%EF%B8%8F-add-cname-file-cname) for documentation.


slondr (approved):
Love it

Issue #745, opened 2020-12-28
Server doesn't respect changes in "default output device" on macOS


kubukoz:
**Description**

The server doesn't change the device playing when the default changes.

**To Reproduce**
1. Play music on built-in speakers
2. Connect headphones or some other extra output
3. Make sure it's now the device in this menu:

<img width="228" alt="image" src="https://user-images.githubusercontent.com/894884/103185697-cff6aa00-48bd-11eb-92aa-401cb2a93da3.png">

4. Sound still plays on speakers

The same happens if you change the default device yourself.

**Expected behavior**
Sound switches to the newly "defaulted" device

**Logs**
<details><summary>Click to show logs</summary>

  

</details>

**Versions (please complete the following information):**

- OS: macOS Catalina 10.15.6 (19G2021)
- Spotifyd: 0.2.24

whitepaperclip:
+1

shatteringlass:
I guess it's a rodio backend bug, rather than related spotifyd. Librespot and ncspot have similar bugs as well. It's worth trying a newer rodio build, including the latest commits.

chazzox:
Hi! Anyone manage to fix this issue? Currently looking around for a fix? @shatteringlass you said to try a more recent rodio build. How would I go about accomplishing this?

shatteringlass:
The issue within rodio was identified to be upstream (https://github.com/RustAudio/rodio/issues/327) and fixed in cpal v.0.14.1, that was merged into rodio only with v0.16.0. 
Currently, spotifyd is still using rodio v0.13.1, so the bug could be solved with a simple PR (https://github.com/Spotifyd/spotifyd/pull/1191).

myInstagramAlternative:
Could this issue be revisited, please? I checked librespot, and rodio is now at version 19.0. Is there anything preventing us from using librespot v0.6.0?

eladyn:
@myInstagramAlternative It has taken some time, but #1317 does exactly this, if you'd be willing to test the changes. (If you're using user+password auth, maybe rather try #1321, since that implements the "new" OAuth flow.)

eladyn:
I'm going to close this as I expect it to be fixed. I don't have access to a macOS machine however, so no easy way to test this.

PR #1191, opened 2023-03-17
Update rodio to 0.17.1 and cpal to 0.15.1

shatteringlass:
Solves #745

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -578,9 +578,9 @@ dependencies = [
 
 [[package]]
 name = "cpal"
-version = "0.13.5"
+version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74117836a5124f3629e4b474eed03e479abaf98988b4bb317e29f08cfe0e4116"
+checksum = "f1241019dec4a73f874bdf6fe3467a8478b47cecaf79435f7acb01f107ab41b1"
 dependencies = [
  "alsa 0.6.0",
  "core-foundation-sys",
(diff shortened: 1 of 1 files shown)


eladyn:
Thank you for the suggestion! Unfortunately, updating that dependency isn't as easy as updating `Cargo.lock`.

```
➜ cargo tree --invert=rodio --all-features
rodio v0.13.1
└── librespot-playback v0.2.0
    ├── librespot-connect v0.2.0
    │   └── spotifyd v0.3.4 (/home/user/spotifyd)
    └── spotifyd v0.3.4 (/home/user/spotifyd)
```

As you can see, we pull in the dependency only indirectly through `librespot-playback`. And even on the current `dev` version, [their `Cargo.toml`](https://github.com/librespot-org/librespot/blob/a211ff94c6c9d11b78964aad91b2a7db1d17d04f/playback/Cargo.toml#L46-L47) requires the older rodio version.

(Compiling your version with `--locked` fails with: `error: the lock file /home/user/spotifyd/Cargo.lock needs to be updated but --locked was passed to prevent this`.)

So what should instead be done is making a PR to `librespot` that updates their `Cargo.toml` and fixes the potential compilation errors that arise from this. :)

eladyn:
Sure, no worries! Regarding your changes to `librespot`, you'll need to upgrade `jack` to “0.5” as well, since otherwise there are dependency conflicts on the `jack-sys` crate between `jack` and `cpal`. But otherwise, I think you could submit that PR. :)

PR #1192, opened 2023-03-19
switch to project domain for controls interface

eladyn:
Since we now have our own domain, we should use that before committing to the one from GitHub before the release.

(The shorter name completely changed the formatting somehow.)


slondr (approved):
epic

PR #1197, opened 2023-03-26
Device Name in Spotify remains "Librespot" ulbi/spotifyd#1

ulbi:


eladyn (commented):
Oh, nice catch! Good that you discovered that before the release. :see_no_evil:

Now that we're at it, at least from my testing, we can remove [this restriction](https://github.com/Spotifyd/spotifyd/blob/08fedd258743801b0e8501c71df26dfaeb84476a/src/config.rs#L763). (and the corresponding [comment](https://github.com/Spotifyd/spotifyd/blob/08fedd258743801b0e8501c71df26dfaeb84476a/docs/src/config/File.md?plain=1#L75) in the docs)

Apart from that, could you please run `cargo fmt` on that, otherwise CI will complain.

ulbi:
> Oh, nice catch! Good that you discovered that before the release. 🙈
> 
> Now that we're at it, at least from my testing, we can remove [this restriction](https://github.com/Spotifyd/spotifyd/blob/08fedd258743801b0e8501c71df26dfaeb84476a/src/config.rs#L763). (and the corresponding [comment](https://github.com/Spotifyd/spotifyd/blob/08fedd258743801b0e8501c71df26dfaeb84476a/docs/src/config/File.md?plain=1#L75) in the docs)
> 
> Apart from that, could you please run `cargo fmt` on that, otherwise CI will complain.

Thanks for the feedback. Did run `cargo fmt` and dropped the space limitations as well. Hopefully it will go through now.
BTW: I did add a log entry for the device_id - need this for monitoring and regularly checking whether the spotifyd is still running in it's docker container.

src/setup.rs
```diff
@@ -90,19 +90,21 @@ pub(crate) fn initial_state(config: config::SpotifydConfig) -> main_loop::MainLo
...
-            .launch()
-            .unwrap();
-        discovery_stream.into()
-    };
+    let credentials_provider =
+        if let Some(credentials) = get_credentials(&cache, &username, &password) {
+            CredentialsProvider::SpotifyCredentials(credentials)
+        } else {
+            info!("no usable credentials found, enabling discovery");
+            info!("device_id {}", session_config.device_id.clone());
```
eladyn, line 98: If we want to add it for everyone, I would change that a bit. Since this log doesn't add much value for normal users, I think it should be debug, so you have to opt in (`--verbose`) to see it.

```suggestion
            debug!("Using device id '{}'", session_config.device_id);
```

docs/src/config/File.md
```diff
@@ -72,7 +72,6 @@ volume_controller = "alsa"  # use softvol for macOS
 on_song_change_hook = "command_to_run_on_playback_events"
 
 # The name that gets displayed under the connect tab on
-# official clients. Spaces are not allowed!
```
eladyn, line 75: Seems like you dropped a bit much there :sweat_smile: (the first two words are still relevant).

eladyn (changes requested):
Two last comments, after that I think we're good to go.

src/setup.rs
ulbi: Done

eladyn (approved):
Looks good and works. Thank you!

PR #1183, opened 2023-03-08
Update dependencies, except librespot

Icelk:
There were no major changes to the code. Tested to function as expected.

PR #1199, opened 2023-03-27
Update tokio

eladyn:
This updates `tokio` which was (I think) the only one that didn't get updated in #1183. Seems to work fine from my testing.


slondr (approved):
awesome

PR #1214, opened 2023-07-16
re-enable credential caching

eladyn:
I just noticed that we disabled credential caching by accident during the `librespot` upgrade.


slondr (approved):
epic

Issue #558, opened 2020-04-05
spotifyd from command line crashes on first connect on LibreElec.


voltagex:
**Description**

Running spotifyd from the command line on a Pi LibreElec install causes it to exit/crash on first connect

**To Reproduce**

1. Download and unpack spotifyd to a writable path on the LE install - I used /storage/bin
2. Run ./spotifyd --no-daemon
3. Play music via e.g. an Android phone.

```
LibreELEC:~/bin # ./spotifyd --no-daemon
No config file specified. Running with default values
No proxy specified
Using software volume controller.
Connecting to AP "gae2-accesspoint-b-vpx4.ap.spotify.com:443"
Authenticated as "voltagex_us" !
Using alsa sink
Country: "AU"
Caught panic with message: Failed to initialize DBus connection: D-Bus error: Unable to autolaunch a dbus-daemon without a $DISPLAY for X11 (org.freedesktop.DBus.Error.NotSupported)
```

if I add DISPLAY=0 to the command line, the crash is

```
Caught panic with message: Failed to initialize DBus connection: D-Bus error: /usr/bin/dbus-launch terminated abnormally with the following error: Autolaunch requested, but X11 support not compiled in.
Cannot continue.
 (org.freedesktop.DBus.Error.Spawn.ExecFailed)
```

**Expected behavior**
Music to play? I'm not sure why an X11 display is required here.,

**Logs**
<details>

```
Caught panic with message: Failed to initialize DBus connection: D-Bus error: Unable to autolaunch a dbus-daemon without a $DISPLAY for X11 (org.freedesktop.DBus.Error.NotSupported)
drop Spirc[0]
Shutting down player thread ...
drop Player[0]
drop Session[0]
drop MercuryManager
shutdown; state=pool::State { lifecycle: Running, num_futures: 0 }
  -> transitioned to shutdown
  -> shutting down workers
dropping I/O source: 3
deregistering handle with poller
dropping I/O source: 0
deregistering handle with poller
dropping I/O source: 1
deregistering handle with poller
dropping I/O source: 2
drop Dispatch
deregistering handle with poller
dropping I/O source: 5
```

</details>

**Versions (please complete the following information):**
 - OS: LibreElec 9.2.1
 - Spotifyd: spotifyd 0.2.24

- cargo:

SimonPersson:
[...]

PR #1215, opened 2023-08-11
Expanded Failed to Initialize D-Bus error message

mas-4:
This is to help people avoid having to dig through issues to find #558

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -166,7 +166,8 @@ async fn create_dbus_server(
         DBusType::Session => connection::new_session_sync(),
         DBusType::System => connection::new_system_sync(),
     }
-    .expect("Failed to initialize DBus connection");
+    .expect("Failed to initialize DBus connection. If you are running headless, \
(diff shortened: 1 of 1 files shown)


mas-4:
I'm not sure. To be honest, it had been over a year since the last time I'd upgraded and accidentally chose the wrong version to upgrade with (I was experiencing frequent disconnections).

If it was a configuration feature to prevent this error, I'd still hit it because I don't have that config flag. So if you decide that's the best route, you'll still need to modify the error message to make it easier, though the error message would be simpler to suggest "Please add disable_mpris=1" to your config if this is expected."

eladyn:
Yeah, modifying the error message would still make sense and I was hoping you might want to do it. 🙈 If you don't have the time or motivation, that's fine as well, and I'll eventually get around to do that some time in thw future. 🙂

mas-4:
I can take a crack at it. I'm unfamiliar with the config system and am a rust noob but I will try this weekend. I like Rust.

PR #1220, opened 2023-08-31
Update dbus_mpris to rspotify 0.12.0

BKasin:
To be honest my main motivation for this is that my `cargo` isn't obeying the .toml and keep pulling rspotify 0.11.7 which breaks mpris, and now that 0.12.0 is released it feels like it makes more sense to update to that than to 0.11.7.

0.11.7 switched completely from std::time::Duration to chrono::Duration, so update our usage accordingly.

0.11.7 also added Collectionyourepisodes to the Type enum, so add it to uri_to_id.

0.12.0 added an optional Market for country code, which for now is just to None in the one spot it was required to be set.

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -30,7 +30,7 @@ use rspotify::{
     prelude::*,
     AuthCodeSpotify, Token as RspotifyToken,
 };
-use std::{collections::HashMap, convert::TryInto, env, pin::Pin, sync::Arc};
+use std::{collections::HashMap, env, pin::Pin, sync::Arc};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 
 pub struct DbusServer {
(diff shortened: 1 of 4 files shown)


src/dbus_mpris.rs
```diff
@@ -296,8 +296,10 @@ async fn create_dbus_server(
                         // MPRIS spec: negative values should be treated as 0
                         let new_pos = new_pos.max(0);
                         if new_pos <= duration as i64 {
-                            let _ =
-                                sp_client.seek_track(new_pos as u32, playback.device.id.as_deref());
+                            let _ = sp_client.seek_track(
+                                Duration::milliseconds(new_pos),
```
eladyn, line 300: Haven't thought about it properly, but now that they're using Durations for `seek_track` (again?), would it maybe be nicer to convert our input (`new_pos`) to a `Duration` and do the addition with the `Duration`s?

eladyn (dismissed):
Looks good to me. If you want, have a look at the one comment, I left.
I currently don't have the means to test the changes, but as long as CI passes that should be fine.

BKasin:
Resolved the lint error from the previous commit, and then changed `new_pos` to a duration (or I guess stopped changing it away from one). Tested by seeking ahead and then back 10 seconds and it works as it should.

mpris

src/dbus_mpris.rs
```diff
@@ -281,23 +281,19 @@ async fn create_dbus_server(
...
-                        .try_into()
-                        .unwrap_or(u32::MAX);
+                        };
 
                         // MPRIS spec: negative values should be treated as 0
-                        let new_pos = new_pos.max(0);
-                        if new_pos <= duration as i64 {
-                            let _ =
-                                sp_client.seek_track(new_pos as u32, playback.device.id.as_deref());
+                        let new_pos = new_pos.max(Duration::zero());
```
eladyn, line 294: Is this line necessary, now that we're using Durations? I'd assume that Durations could never be negative by the type itself?

eladyn (commented):
That looks good to me, just one last comment. Then we should be good to go.

src/dbus_mpris.rs
BKasin: std::Duration only allows positive values, but chrono::Duration allows negative and is what rspotify uses for everything. So I think it is still useful to make sure `new_pos` is >= 0? But I could be wrong.

eladyn: Oh, yeah. It's actually quite important that they support negative values for us to be able to seek backwards. 😅

eladyn:
Looking at CI, it seems like we might need a MSRV bump? Would you be able to update that in the CI as well as in the Cargo.toml?

BKasin:
It looks like the version of `time` required by `cookie`, `cookie_store`, and `syslog` went from `0.3.20` to `0.3.28` and that's what caused the bump.

eladyn (approved):
Thank you!

PR #1229, opened 2023-10-15
fix alsa mixer calculations

### LINKED ISSUE #1222: Volume Controllers behave weirdly
**Description**
softvol: changes it just for spotify and properly from 0 to 100 (I think), but doesnt show these changes in system volume mixer
alsa: Sets 0 volume to be around 80% actual volume so its very loud when I try to increase the volume from my phone through spotify.
alsa_linear: Properly sets it from 0 to 100, but sets system volume and not just "spotifyd" and resets it to 0 everytime I pause and play
none: Doesnt let me change it from the phone

**To Reproduce**
1. Install spotifyd
2. try all different volume controllers

**Expected behavior**
All should show spotifyd's volume on a separate slider in mixer and properly set the volume from 0 to 100

**Logs**
I can't provide logs for all four controllers

- OS: Fedora 38
- Spotifyd: 0.3.5
- cargo: 1.72.0

eladyn:
fixes #1222

I hope, this fixes the issues mentioned with `alsa` and `alsa_linear`. Please note that one of them will behave weirdly on your system depending on, whether your alsa control is linear or not. (At least that is what I assume based on my experience...) However, they should both be able to reach 0% and not be reset when hitting play.

Any testing is appreciated. (maybe @Arbee4ever?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Credential caching has been re-enabled. ([#1214])
 
+### Fixed
+- alsa mixer: volume calculation has been fixed ([#1229])
+
 [#1214]: https://github.com/Spotifyd/spotifyd/pull/1214
+[#1229]: https://github.com/Spotifyd/spotifyd/pull/1229
 
 ## [0.3.5]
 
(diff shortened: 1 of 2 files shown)


Ionshard:
I'd be happy to test this, but I run spotifyd on a Raspberry Pi (2B), which is currently suffering from this issue (and have switched to using softvol) and I am not sure how to get this on there. I setup by just downloading the Armv6 binary. Do I just compile it on the Pi?

PR #1247, opened 2023-12-13
Explain that credentials from keyring can't be accessed in system service

sepastian:
This took me a while to figure out, so I created a PR updating the docs with something that's missing, IMO.

Systemd allows running spotifyd as a system service, or as a user service.

When running as a system service, accessing login credentials stored in a user's keyring is not possible.

When using the user's keyring to store login credentials, as described in the docs, make sure to also run the spotifyd as a user service.


eladyn (approved):
That seems like a sensible addition and might reduce the confusion. Thank you!

PR #1253, opened 2023-12-27
Recognise OpenBSD to unbreak build

klemensn:
```
  error[E0425]: cannot find value `get_shell_ffi` in this scope
    --> src/utils.rs:69:48
```

Otherwise spotifyd-0.3.5 works great on OpenBSD/amd64 7.4-current.


eladyn (approved):
Thank you! I didn't try building and running on an OpenBSD system yet, but changes look fine!

klemensn:
> Thank you! I didn't try building and running on an OpenBSD system yet, but changes look fine!

Not sure if `cargo build` from a checkout just works, but the official port/package sure does, which is what I have been doing exclusively.

PR #1254, opened 2023-12-31
Bump keyring to latest release 2.3.0 to add OpenBSD support

klemensn:
After 2.2.0 I submitted https://github.com/hwchen/keyring-rs/pull/156 which turned into 2.3.0 today.

OpenBSD's port currently needs patches, see PR for details.

I have only tested spotifyd's 0.3.5 release with this keyring bumped 2.3.0 instead of patching 2.0.5 and it works.

Done with `cargo update keyring`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1460,9 +1460,9 @@ dependencies = [
 
 [[package]]
 name = "keyring"
-version = "2.0.5"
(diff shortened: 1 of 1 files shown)


eladyn (approved):
Thanks!

PR #1256, opened 2024-01-05
Update ureq, rustls and sct to bump ring>=0.17

klemensn:
On OpenBSD, ring 0.16 a) needs patches to comply with execute-only memory policies and b) does not build on sparc64, riscv64 and powerpc64.

Works for me in combination with spotify-qt on OpenBSD/amd64 7.4-current.

Done with `cargo update ureq rustls sct`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -549,9 +549,9 @@ dependencies = [
 
 [[package]]
 name = "cookie"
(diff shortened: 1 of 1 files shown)


eladyn (approved):
Looks good! Thank you for all your work on the BSD support!

PR #1257, opened 2024-01-05
Default to PortAudio backend on non-Linux

klemensn:
- macOS homebrew[0], NetBSD pkgsrc[1], FreeBSD[2] and OpenBSD[3] ports diable default features and do not reenable "alsa_backend" by default
- macOS CI builds do not use it
- Windows has no ALSA

They all do, however, enable "portaudio_backend" to use PortAudio by default.

Reflect that such that the aforementioned four platforms/operating systems do not have to override features whilst making ALSA Linux-only by default.

I am not sure what backend is used on Windows by default, but PortAudio aught to work there;  certainly better than ALSA.

0: https://github.com/Homebrew/homebrew-core/blob/master/Formula/s/spotifyd.rb
1: http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/audio/spotifyd/Makefile?rev=1.2&content-type=text/x-cvsweb-markup
2: https://cgit.freebsd.org/ports/tree/audio/spotifyd/Makefile
3: http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/audio/spotifyd/Makefile?rev=1.6&content-type=text/x-cvsweb-markup


eladyn:
Thank you for that contribution! Unfortunately, the target specific features [don't seem to be working](https://github.com/Spotifyd/spotifyd/actions/runs/7427072683/job/20212075744?pr=1257#step:6:8). That's probably a missing cargo feature, however: https://github.com/rust-lang/cargo/issues/1197.

Maybe a different solution would be to use `rodio` by default which is apparently designed to be a cross platform audio playback library? Not sure, how it works on things different than linux though.

klemensn:
> Thank you for that contribution! Unfortunately, the target specific features [don't seem to be working](https://github.com/Spotifyd/spotifyd/actions/runs/7427072683/job/20212075744?pr=1257#step:6:8). That's probably a missing cargo feature, however: [rust-lang/cargo#1197](https://github.com/rust-lang/cargo/issues/1197).

I see, that doesn't fly.
_/me is a rust/cargo rookie_

> Maybe a different solution would be to use `rodio` by default which is apparently designed to be a cross platform audio playback library? Not sure, how it works on things different than linux though.

First and last time I tried, rodio wouldn't build as-is on OpenBSD;  I can look into that, but only on my platform and either way, I have no experience with rodio (haven't even heard about it), whereas I know that PortAudio pretty much just works
on all supported platforms.

klemensn:
For reference:  rodio uses cpal, so it all boils down to https://github.com/RustAudio/cpal/pull/493 from a few years ago.
psst (Rust GUI Spotify client) also uses cpal -- building was one issue, working playback was another, it is all very much WIP on OpenBSD, still.

PR #1264, opened 2024-02-04
Use pledge(2) on OpenBSD to restrict system calls

klemensn:
Limit the attack surface of spotifyd, an internet facing network daemon with read/write filesystem access and options to execute arbitrary commands.

Most importantly, prevent fork(2)/execve(2) unless `onevent` is used.

While the set of runtime promises retains full read/write filesystem as well as network access (for now), it does exclude a number of groups of unused system calls -- the manual[0] for details.

OpenBSD's official package has been shipping this patch for a month by now without any regressions or reports of breakage.

0: https://man.openbsd.org/pledge.2

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,7 +92,7 @@ fn main() -> eyre::Result<()> {
     #[cfg(target_os = "openbsd")]
     pledge(
         "stdio rpath wpath cpath inet mcast flock chown unix dns proc exec audio",
-        None
+        None,
     )
     .unwrap();
 
@@ -199,7 +199,7 @@ fn main() -> eyre::Result<()> {
         if internal_config.onevent.is_some() {
             pledge(
                 "stdio rpath wpath cpath inet mcast unix dns proc exec audio",
-                None
+                None,
             )
             .unwrap();
         } else {
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,8 @@ use log::{info, trace, LevelFilter};
 use std::fs;
 use structopt::StructOpt;
 use tokio::runtime::Runtime;
+#[cfg(target_os = "openbsd")]
+use pledge::pledge;
 
 #[cfg(feature = "alsa_backend")]
 mod alsa_mixer;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -85,6 +87,15 @@ fn setup_logger(log_target: LogTarget, verbose: bool) -> eyre::Result<()> {
...
 
 fn main() -> eyre::Result<()> {
+    // Start with superset of all potentially required promises.
+    // Drop later after CLI arguments and configuration files were parsed.
+    #[cfg(target_os = "openbsd")]
+    pledge(
+        "stdio rpath wpath cpath inet mcast flock chown unix dns proc exec audio",
+        None,
+    )
+    .unwrap();
```
eladyn, line 97: I'm not sure, what the scenarios are, where this fails, but from a user's perspective it might be good to include a short explanation what failed here? Like "failed to restrict service permissions via pledge" or something similar?

eladyn (approved):
As with the other PRs, I didn't test it myself, but since it has been out in the wild in the form of a patch already, this looks good to me. (And thanks for the well documented code!)

src/main.rs
klemensn: https://man.openbsd.org/pledge.2#ERRORS
```
pledge() will fail if:

EFAULT    promises or execpromises points outside the process's allocated address space.
EINVAL    promises is malformed or contains invalid keywords.
EPERM     This process is attempting to increase permissions.
```
Promises are literal values, so the first two won't happen, leaving just `EPERM`.

In any event, first thing to do (upon pledge(2) failure) on OpenBSD would be to run the program through [ktrace(1)](https://man.openbsd.org/ktrace.1) which gathers all relevant information.

`unwrap()` seems enough, is used in the pledge crate's own example and somewhat matches the usual C idiom:
```
if (pledge(..., NULL) == -1)
    err(1, "pledge");
```

eladyn: That sounds very good. Thank you for the clarification.

PR #1279, opened 2024-04-20
Add documentation for D-Bus control

danielmcmillan:
Adds some documentation around the MPRIS support and the additional nonstandard D-Bus interface, which doesn't seem to be available elsewhere outside of Github issues and source code.


eladyn (approved):
Amazing, thank you!

PR #1281, opened 2024-04-29
Add support for `rodiojack` backend.

agdphd:
The underlying `librespot` library already supports the JACK audio backend via `rodio`. Enabling it in `spotifyd` turns out to be dead simple--just needed to add the config enum elements to pass it through!

NOTE: I have tested this on my machine running Debian testing ("trixie"). The documentation for Fedora is based on the upstream docs for librespot: https://github.com/librespot-org/librespot/blob/329a1b16074af4d1a8abe2ed017b5d4a4a070ea1/COMPILING.md#audio-library-dependencies

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -25,7 +25,9 @@ const CONFIG_FILE_NAME: &str = "spotifyd.conf";
     feature = "pulseaudio_backend",
     feature = "portaudio_backend",
     feature = "alsa_backend",
-    feature = "rodio_backend"
+    feature = "rodio_backend",
+    feature = "rodiojack_backend",
+    feature = "jackaudio_backend",
 )))]
 compile_error!("At least one of the backend features is required!");
 static BACKEND_VALUES: &[&str] = &[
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -24,7 +24,9 @@ const CONFIG_FILE_NAME: &str = "spotifyd.conf";
     feature = "pulseaudio_backend",
     feature = "portaudio_backend",
     feature = "alsa_backend",
-    feature = "rodio_backend"
+    feature = "rodio_backend",
+    feature = "rodiojack_backend",
+    feature = "jackaudio_backend",
```
eladyn, line 29: I think, this might be a leftover from your testing? As this feature does not actually exist in the `Cargo.toml`.

src/config.rs
```diff
@@ -36,6 +38,10 @@ static BACKEND_VALUES: &[&str] = &[
     "portaudio",
     #[cfg(feature = "rodio_backend")]
     "rodio",
+    #[cfg(feature = "rodiojack_backend")]
+    "rodiojack",
+    #[cfg(feature = "jackaudio_backend")]
+    "jackaudio",
```
eladyn, line 44: Same here probably?

eladyn (changes requested):
Thank you for this PR! Only some small comments from my side.

eladyn:
And I fixed the clippy lints in #1282, so if could rebase your changes, that would be nice as well

agdphd:
Thanks for your review! Removed the spurious `jackaudio` references and rebased to current `HEAD`.

eladyn (approved):
Thank you and sorry for taking so long for this simple PR!

PR #1284, opened 2024-05-13
add pipe as a backend

### LINKED ISSUE #354: Bring back pipe backend
#283 removed the pipe backend during the Great Refactoring (TM pending). Might be a good idea to bring it back at some point.

Thinks to take care of:
- [ ] Logging to stdout would interfere with this feature. So if `--backend pipe` is specified and the daemon runs with the `--no-daemon` flag, all logging would have to happen on stderr instead.
- [ ] Add the new backend to the `Backend` enum.
- [ ] Add the needed code to `setup.rs` (I think this is were the backend logic lies).

"Fixes" #256 by basically providing the functionality in the first place.

zemar1325:
add pipe as a backend, which sends audio to stdout or the file handle given to --device

Fixes #354


eladyn (approved):
Sorry for taking so long for this quite simple PR. Looks good to me though.

eladyn (approved):

Issue #1053, opened 2022-02-06
does not start on raspberry pi 3 b+ with latest raspberry pi os 64 bit


meeeeeeeeeeeeeeee:
using systemctl --user start spotifyd.service does nothing

starting spotifyd with ./spotifyd --no-daemon I get ./spotifyd: error while loading shared libraries: libsystemd.so.0: cannot open shared object file: No such file or directory

checking to see if that file exists says the one for aarch64 is there but the one for armhf is missing

pi@raspberry-pi:~ $ ls -l /lib/aarch64-linux-gnu/libsystemd.so.0*
lrwxrwxrwx 1 root root     20 Jul 13  2021 /lib/aarch64-linux-gnu/libsystemd.so.0 -> libsystemd.so.0.30.0
-rw-r--r-- 1 root root 742816 Jul 13  2021 /lib/aarch64-linux-gnu/libsystemd.so.0.30.0
pi@raspberry-pi:~ $ ls -l /lib/arm-linux-gnueabihf/libsystemd.so.0*
ls: cannot access '/lib/arm-linux-gnueabihf/libsystemd.so.0*': No such file or directory

trying to install the armhf version of systemd fails because it wants to remove the 64 bit version and can't

a fix for this is to have precompiled binaries for armhf and aarch64

eladyn:
Hi!

I have not much experience in this matter, but I found out that only the `-default` and `-full` [releases](https://github.com/Spotifyd/spotifyd/releases/tag/v0.3.3) are linking to `libsystemd` (AFAICT). So you might get further by trying the `-slim` version.

Apart from that, I don't know how feasible it is to execute binaries compiled for `armhf` on `aarch64`. However, you can of course [build `spotifyd` yourself](https://spotifyd.github.io/spotifyd/installation/Ubuntu.html) directly on the platform.

Hope this helps!

waldeckmatheus:
Hello, here i found the same problem, i wasn't able to do so, downloaded spotifyd but not happens when running: spotifyd --no-daemon

I "solved" this by doing the following:

Compiled it with aarch64 (arm64v8), in a RPI 3 model B, OS release Debian GNU/Linux 11 (bullseye).

if need to, you may try the following: 
Before compile: installation of libs; increase swap from 100mb to 1024mb.

# Installation of libraries
sudo apt install libasound2-dev libssl-dev libpulse-dev libdbus-1-dev -y
sudo apt install cargo -y

[...]

PR #1304, opened 2024-09-26
Guide for cross compiling using docker

mietzen:
Added a guide for cross compiling using docker for e.g. building arm64 as described in #1053


eladyn (changes requested):
That looks great, thanks!

Since this is a new file in the docs folder, a link in `SUMMARY.md` would have to be added.

You can also check, if everything looks good by  building the docs with `mdbook`.

mietzen:
I added the file to `SUMMARY.md` and tested it with `mdbook`, everything looked good 👍

eladyn (approved):
Amazing! Thank you.

mietzen:
@eladyn I had a look at your CD Pipeline. It could be simplified using this approach. Would you be open / interested to accept a PR for your CI/CD Pipelines?

eladyn:
Yes, absolutely! I was thinking that too, but currently our pipeline is really terrible, so I didn't want to put that burden on you. 😅 But if you're motivated that would be great!

PR #1312, opened 2024-11-27
Show supported features and audio backend in --version

### LINKED ISSUE #1255: Show release build type in --version (slim, full, default)
Currently, `spotifyd --version` only shows:

```
spotifyd 0.3.5
```

it would be nice if it also listed if this is an official release, and which options it was (and wasn't compiled with), so for example

```
spotifyd 0.3.5
slim
NO mpris support
...
```

antomfdez:
Fixes #1255

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ librespot-connect = { version = "0.4" }
 toml = "0.7"
 color-eyre = "0.6"
 directories = "5.0.1"
+once_cell = "1.20.2"
 
 [target."cfg(unix)".dependencies]
 daemonize = "0.5"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2197,9 +2197,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -36,6 +36,7 @@ librespot-connect = { version = "0.4" }
 toml = "0.7"
 color-eyre = "0.6"
 directories = "5.0.1"
+once_cell = "1.20.2"
```
eladyn, line 39: In theory, since recently, the standard library also has support for those Lazy cells e.g. with `std::cell::LazyCell`. But to make use of that, we would need a MSRV bump, so this would also be fine for now! Whatever suits you best. :)

eladyn (changes requested):
Thank you and sorry for the long delay! I made some comments, but it looks like a good approach I think.

antomfdez:
> Thank you and sorry for the long delay! I made some comments, but it looks like a good approach I think.

I'm gonna take a look the next few days to see if I can fix it

PR #1314, opened 2024-12-16
Enable build on android

thesebas:
This enables build in Termux on android

As a "proof" I'm providing a screenshot from the phone that it actually builds (with some features restricted) and it runs.

Bulit with:
```
cargo build --release --no-default-features --features pulseaudio_backend
```

and run with 
```
DISPLAY=0 ./target/release/spotifyd --no-daemon
```

Screenshot_20241216-134824

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,7 +1,12 @@
 use log::trace;
 use std::env;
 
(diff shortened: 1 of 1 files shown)


eladyn (approved):
Ooh, that's nice! Thank you.

PR #1315, opened 2024-12-19
Add definition of `Seeked` MPRIS signal

dotandl:
This PR adds a lacking definition of the `Seeked` signal on the `org.mpris.MediaPlayer2.Player` DBus interface.

While the signal is working and `dbus-monitor` can listen to it, it is not visible in `d-feet` and causes problems with some DBus client libraries, e.g. [node's `dbus-next`](dbusjs/node-dbus-next).

### DIFF
--- a/src/dbus_mpris.rs
+++ b/src/dbus_mpris.rs
@@ -226,10 +226,16 @@ async fn create_dbus_server(
             .get(|_, _| Ok(Vec::<String>::new()));
     });
 
+    let mut on_seeked: Option<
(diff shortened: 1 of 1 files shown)


eladyn:
Thank you, that sounds like a good idea. Since I did some major rewrite of the `dbus_mpris` part recently, I adopted your changes and would merge it as part of that PR (#1317). Does that sound/look good?

dotandl:
Sure, that sounds great.

PR #1316, opened 2024-12-19
Add new device types

dotandl:
`spotifyd` was lacking new device types (such as `automobile`, `gameconsole` etc.) introduced some time ago in `librespot`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -146,28 +146,28 @@ static DEVICETYPE_VALUES: &[&str] = &[
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, StructOpt)]
 #[serde(rename_all = "snake_case")]
 pub enum DeviceType {
-    Unknown = 0,
-    Computer = 1,
-    Tablet = 2,
-    Smartphone = 3,
-    Speaker = 4,
+    Unknown,
+    Computer,
+    Tablet,
+    Smartphone,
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -150,6 +158,16 @@ pub enum DeviceType {
...
+    GameConsole = 9,
+    CastAudio = 10,
+    CastVideo = 11,
+    Automobile = 12,
+    Smartwatch = 13,
+    Chromebook = 14,
+    UnknownSpotify = 15,
+    CarThing = 16,
+    Observer = 17,
+    HomeThing = 18,
```
eladyn, line 170: I'm a bit confused, why we need those numbers in the enum (not only in the changes you introduced, but also above). If you have the time, we could maybe get rid of them? But surely not necessary for this PR to be mergeable.

eladyn (dismissed):
Thank you! Some minor comment that you can address, if you want, but otherwise I'd be ready to merge.

src/config.rs
dotandl: Well, to be honest I wasn't sure why these enum entries have been explicitly assigned a value either. I've done the same to make the code more consistent but now I'm sure this makes no sense (this is the only enum constructed in such way).

I'll push a commit with the values removed in a few minutes.

eladyn (approved):
Looks good, awesome!

Issue #1359, opened 2025-08-11
Spotifyd not switching sessions when accessed by multiple accounts


todorangrg:
**Description**
Switching sessions on different account connects doesn't seem to work reliably for me. Most of the time it times out. Sometimes it switches after 10-20s wait. For what it's worth, the 2 accounts are part of the same family plan. I did not set up `credentials.json` for any account, it's zeroconf only.

My understanding from reading previous issues is that this should be working nicely.

**To Reproduce**
1. Start spotifyd (clear cache, no credentials.json in oauth nor zeroconf) 
2. Connect with account A, play a bit, pause, cancel Spotify connect -- choose back device playback)
     **-- everything works as expected**
4. Connect with account B, play a bit, pause, cancel Spotify connect -- choose back device playback)
     **-- session switch sometimes happens, but it needs 10-20s to start new session**
6. Connect with account A
     **-- doesn't connect at all, even multiple retries, even 30-40min later**
8. Clean re-starting spotifyd (removing cache) connects no-problem.

**Expected behavior**
Account switching happens seamlessly, similar with the first-time connect of a freshly started spotifyd instance.

**Logs**
<details><summary>Click to show logs</summary>

```
2025-08-10T23:24:37.459772654Z [DEBUG librespot_core::session] Session strong=5 weak=7
2025-08-10T23:24:49.802180899Z [DEBUG libmdns::fsm] received question: IN _googlecast._tcp.local
2025-08-10T23:24:49.802482784Z [DEBUG libmdns::fsm] received question: IN _CC32E753._sub._googlecast._tcp.local
2025-08-10T23:24:49.802529452Z [DEBUG libmdns::fsm] received question: IN _googlecast._tcp.local
2025-08-10T23:24:49.802560953Z [DEBUG libmdns::fsm] received question: IN _CC32E753._sub._googlecast._tcp.local
2025-08-10T23:24:50.083398218Z [DEBUG libmdns::fsm] received question: IN _spotify-connect._tcp.local
2025-08-10T23:24:50.153441935Z [DEBUG librespot_core::mercury] unknown subscription uri=hm://remote/user/my_user_a_username/
2025-08-10T23:24:50.153632691Z [DEBUG librespot_core::session] could not dispatch command: Service unavailable { error handling Mercury response: MercuryResponse { uri: "hm://remote/user/my_user_a_username/", status_code: 200, payload: [[bunch of numbers]] } }
[...]

PR #1361, opened 2025-08-21
fix for #1359

todorangrg:
this only fixed the issue I was having in https://github.com/Spotifyd/spotifyd/issues/1359. Now user-switch works reliably.

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -205,6 +205,7 @@ impl MainLoop {
                     .set_session(shared_spirc.clone(), connection.session)
                 {
                     let _ = shared_spirc.shutdown();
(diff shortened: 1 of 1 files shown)


eladyn:
@todorangrg The log you pasted only shows two warning messages (which were already present on master and I fixed them now). I just rebased this branch onto the latest changes, so maybe you can try again? (Compiling might take a little...)

eladyn (changes requested):
Sorry for taking so long to review this! Looks good though, thank you! There are a few other places, where `shared_spirc.shutdown()` is called. Would it be possible to await the `spirc_task` there as well?

eladyn (approved):

PR #1362, opened 2025-08-25
chore: bump librespot to 0.7.1

fabienjuif:
Just a quick update to make it works at home.
Feel free to edit/discard.


src/setup.rs
```diff
@@ -44,7 +44,8 @@ pub(crate) fn initial_state(
             }
             _ => {
                 info!("Using software volume controller.");
-                Arc::new(mixer::softmixer::SoftMixer::open(MixerConfig::default()))
+                let soft_mixer = mixer::softmixer::SoftMixer::open(MixerConfig::default()).expect("softmixer should initialize");
```
eladyn, line 47: Check that this will never panic.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ env_logger = "0.11"
 [features]
 alsa_backend = ["librespot-playback/alsa-backend", "dep:alsa"]
 dbus_mpris = ["dep:dbus", "dep:dbus-tokio", "dep:dbus-crossroads"]
-default = ["alsa_backend"]
+default = ["alsa_backend", "pulseaudio_backend", "dbus_mpris"]
```
eladyn, line 64: Where does this change come from?

src/setup.rs
```diff
@@ -135,7 +134,7 @@ pub(crate) fn initial_state(
         player_config,
         backend,
         initial_volume: config.initial_volume,
-        has_volume_ctrl,
+        disable_volume: false,
```
eladyn, line 137: I don't think, this is what we want. The previous flag will be ignored this way.

eladyn (changes requested):
Thanks a lot for the fixes! I added a lot of comments, but don't worry, I can just fix them myself, since I have ignored this PR for too long now. So these are more of a todo list before merging this PR.

clxxiii:
Somewhat new to open source here:
What's the best way for me to keep working on the hotfix branch while keeping the discussion on this PR? I'd love to help out getting this fixed where I can, and I have the time to be able to help out!
I can also just start on my own fork and open another PR if that makes more sense

PR #1372, opened 2025-10-22
add rpm building

sckova:


eladyn (approved):
Thank you!
(And I just noticed that we haven't been keeping the Contributors.md up to date at all, maybe just dropping it or re-enabling it would be good.)

Issue #1389, opened 2026-01-06
Cannot open shared object error libssl.so.1.1 - v0.4.2 aarch64-default


JimZipCode:
**Description**
On Raspberry Pi OS for Pi4, ./spotifyd --version throws error:

`./spotifyd: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

**To Reproduce**
1. Used wget to download v0.4.2/spotifyd-linux-aarch64-default.tar.gz
2. tar xf 
3. chmod +x
4. chown root:root
5. ./spotifyd --version

**Versions**

- OS: Raspberry Pi OS; Debian 13.2 (trixie); 
- uname -a output: 
`Linux pi4serv 6.12.47+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.12.47-1+rpt1 (2025-09-16) aarch64` 
- Spotifyd: v0.4.2 linux-aarch64-default

**Additional context**
Running `find / -type f -name "libssl.so*"` produces this list:

- /usr/lib/aarch64-linux-gnu/libssl.so.3
- /opt/emby-server/lib/libssl.so.3

eladyn:
Thank you for the report. The libssl situation is indeed a little unfortunate.
For aarch64, this should be fixed by https://github.com/Spotifyd/spotifyd/pull/1387. Maybe you can try one of the binaries that are automatically built from the master branch: https://github.com/Spotifyd/spotifyd/actions/workflows/cd.yml. (e.g. one of [these](https://github.com/Spotifyd/spotifyd/actions/runs/20684935673))

sorinpohontu:
Nope, I have test the artifact spotifyd-linux-armv7-slim and spotifyd-linux-armv7-full, same error:

`/usr/bin/spotifyd: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

eladyn:
On `armv7` this is indeed unfortunately not yet resolved, sorry about that. :/ (The issue above was about `aarch64` originally, which should hopefully be fixed now.)

pmcgaley:
The CD builds linked above seem to have expired, and looks like CD is configured to run only on merge to `master`, which hasn't happened since February. Any chance of an official release to support `aarch64`? Or even someone triggering CD to get a new build out?

MichaIng:
[...]

PR #1400, opened 2026-03-10
ci: add armv6 (Raspberry Pi B/Zero) build support

b0bbywan:
Spotifyd dropped support for armhf after v0.3.5. This intends to bring it back

Cross-compilation via `cross` doesn't work for armv6 due to C dependencies (alsa, dbus, pulseaudio) and bindgen (aws-lc-rs). Instead, build natively under QEMU using a RaspiOS container, forcing the correct rustup host triple (arm-unknown-linux-gnueabihf) since QEMU reports armv7 by default.

- Add .github/Dockerfile.armv6 for QEMU-native builds
- Add armv6 to the CD matrix (linux only, slim/default/full)
- Use docker buildx with GHA layer cache in the CD workflow
- Add dist/ to .gitignore for local build

I've successfully tested the slim and full builds on Raspberry B and B+. The first build time is quite long (more than one hour) due to QEMU virtualization. It uses a rust cache to speedup a bit. We could also push the build image on ghcr to  gain a little time. 

Let me know if that's ok with you or if I need to make some change.

Thanks for spotifyd !

### DIFF
--- a/.github/Dockerfile.armv6
+++ b/.github/Dockerfile.armv6
@@ -1,0 +1,74 @@
+# syntax=docker/dockerfile:1
+
+# ── Builder base images ────────────────────────────────────────────────────────
+# amd64: native build on Debian Trixie
+# arm64/arm: RaspiOS (Trixie), runs natively via QEMU binfmt_misc —
+#   no cross-compiler needed, even for bindgen
+FROM debian:trixie-slim       AS base-amd64
+FROM vascoguita/raspios:arm64 AS base-arm64
+FROM vascoguita/raspios:armhf AS base-arm
+
+# ── Builder ───────────────────────────────────────────────────────────────────
+ARG TARGETARCH
+FROM base-${TARGETARCH} AS builder
+
+ARG TARGETARCH
+ARG TARGETVARIANT
+# Default: slim build with alsa only
+ARG FEATURES="--no-default-features --features alsa_backend"
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl \
+    ca-certificates \
+    gcc \
+    libc6-dev \
+    # required by aws-lc-rs (bindgen) on non-x86_64/aarch64 targets
+    libclang-dev \
+    cmake \
+    pkg-config \
+    # alsa_backend + base
+    libasound2-dev \
+    libssl-dev \
+    # dbus_mpris
+    libdbus-1-dev \
+    # pulseaudio_backend
+    libpulse-dev \
+    # rodiojack_backend
+    libjack-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Force the correct rustup host triple.
+# QEMU reports armv7 even inside an armv6 container — override manually.
+RUN set -eux; \
+    case "${TARGETARCH}${TARGETVARIANT}" in \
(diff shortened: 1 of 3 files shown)


b0bbywan:
https://github.com/Spotifyd/spotifyd/issues/1389 this issue could be fixed in that PR, let me know !

b0bbywan:
Hello,

Just so you know, I'm working on an open-source Volumio alternative (https://beta.odio.love) that ships spotifyd as a dependency.

I noticed all ARM builds were broken due to OpenSSL 1.1 linkage (issue #1389), so I switched to native QEMU-based builds inside RaspiOS containers in my fork. This produces binaries correctly linked against OpenSSL 3, and I've tested them on real hardware (armv6, armv7, aarch64). I also publish them through my own APT repo with proper Debian packaging.

I submitted PR #1400 with the CI changes for armv6. I'd love to work together on extending this approach to all ARM targets and shipping working binaries on modern distros.

Happy to help however I can, let me know if you'd like to discuss the approach.

robertaistleitner:
@b0bbywan thanks for your fixes on making this work on older raspberries again. I just updated my Raspberry 2B to Trixie and also upgraded to the latest spotifyd which no longer works in 0.4.2 - your build 0.4.4 (armv7-slim) from https://github.com/b0bbywan/spotifyd/releases works perfectly fine!

.github/Dockerfile.armv6
```diff
@@ -0,0 +1,74 @@
+# syntax=docker/dockerfile:1
+
+# ── Builder base images ────────────────────────────────────────────────────────
+# amd64: native build on Debian Trixie
+# arm64/arm: RaspiOS (Trixie), runs natively via QEMU binfmt_misc —
+#   no cross-compiler needed, even for bindgen
+FROM debian:trixie-slim       AS base-amd64
+FROM vascoguita/raspios:arm64 AS base-arm64
```
MichaIng, line 8: Why Raspberry Pi OS for arm64? I'd use official Debian slim really for everything but armv6. Those are marked as official Docker images, assured to remain stable, updated, slim etc. Even if Raspberry Pi OS uses the Debian repository for arm64, it adds their own Raspberry Pi repo for kernel and also some other/overlapping library packages on top, potentially causing conflicts on non RPi systems, overhead in every case, and vascoguita is not a trusted publisher.

MichaIng:
Note that #1387 fixed the libssl1.1 issue for `aarch64`/`arm64` already. #1420 would do so for `armv7l`/`armhf`, still using `cross-rs`, but its latest commit and edge images, which use Ubuntu 24.04 now, i.e. libssl3.

So building on Raspberry Pi OS and via QEMU remains necessary for `armv6hf` only, not supported by the `cross-rs` Ubuntu base images. It might have benefits to drop `cross-rs` for own QEMU-emulated builds, so everything is under local control, instead of relying on `cross-rs` which took years to finally bump Ubuntu base images to a usable version. And this is still no tagged release. But QEMU-emulated builds are significantly slower compared to cross-compiling. Total CD duration [above 30 minutes](https://github.com/b0bbywan/spotifyd/actions/runs/23553734369) vs [below 10 minutes](https://github.com/MichaIng/spotifyd/actions/runs/28552458687) with cross-compiling. A single build with `cross-rs` takes ~5 minutes. 30 minutes seems totally acceptable though, I expected it to take longer.

PR #1408, opened 2026-04-16
WIP: making sure a single daemon instance is being executed

zanella:
Hi,

I often forget if I have a running spotifyd instance or not, and also having a single daemon running feels more correct.

Assuming this is a feature interesting enough to be merged, questions: 

1 - Would the path for the lock file need to exist in the cfg ? And be configurable ? 
2 - Would it be necessary to have another cmd line option? e.g. --single-daemon
3 - It's the first time I write Rust, still need to figure out how to add a test for this.
3 - I don't have a Windows machine, have no idea how to test it there


zanella:
There's already a PID lock file logic in the code, no need for extra code, just an MR clarifying how to use it.

